PR #592, opened 2025-11-03
human-in-the-loop with Strands + Bedrock 

MohTahsin:
Example using Strands built in human in the loop capabilities, with Runtime's session management capabilities. Using "handoff_to_user" within the Agent tools without the need for external hooks.


MohTahsin:
moving to strands repo

PR #1150, opened 2026-03-20
Update demo

brnaba-aws:
Added header condition on ALB
Added s3 log access on ALB
Added audience and scopes to ACG
Added VPCendpoint with resource based policy
Added path based routing


satveerkhurpa (approved):
Updates reviewed

PR #1557, opened 2026-05-22
feat(payments): add Pay for Data use case (01-features)

joshuamarksmith:
Script-based variant of the Heurist finance agent for the `01-features/08-agents-that-transact/02-use-cases/` tree.

Single Python file (`pay_for_data.py`) handles the full lifecycle: provision embedded wallet, deploy to AgentCore Runtime, create payment session, invoke the agent, and verify spend. Designed for AI-agent-driven execution rather than manual notebook cells.

## What it does

A finance research agent calls paid Heurist x402 endpoints for live market data, analyzes results with AgentCore Code Interpreter, and exports charts/reports to S3 — settling USDC on Base mainnet via `AgentCorePaymentsPlugin`.

## Layout

```
pay-for-data/
├── pay_for_data.py          # Full lifecycle script (provision → deploy → invoke → verify)
├── requirements.txt         # Client-side Python deps
├── .env.sample              # Environment template
├── .gitignore
├── README.md
└── agent/                   # Ships in the Runtime container
    ├── main.py              # BedrockAgentCoreApp entry point
    ├── catalog.py           # Heurist registry loader
    ├── config.py            # Environment config
    ├── sync_registry.py     # Catalog cache refresh CLI
    ├── requirements.txt     # Container deps
    └── Dockerfile
```

Mirrors the format of `pay-for-content-browser-use/` in the same directory.


TateLyman:
Ran a source pass on the script-based Pay for Data sample and found one lifecycle blocker in the catalog sync path.

`pay_for_data.py` Step 5 does:

```py
sys.path.insert(0, "agent")
from catalog import fetch_live_catalog

fetch_live_catalog()
```

But `agent/catalog.py#fetch_live_catalog()` calls `get_config()` just to read `heurist_catalog_url`, and `agent/config.py#get_config()` requires:

```py
PAYMENT_MANAGER_ARN
PAYMENT_SESSION_ID
PAYMENT_INSTRUMENT_ID
```

At Step 5, a payment session has not been created yet; the session is created later in Step 6. The `.env.sample` also tells users to save `MANAGER_ARN`, `PAYMENT_CONNECTOR_ID`, and `PAYMENT_INSTRUMENT_ID`, not `PAYMENT_MANAGER_ARN` / `PAYMENT_SESSION_ID`. So the full lifecycle script can fail before deploy while trying to refresh a public catalog.

I reproduced the failure without AWS, live Heurist, wallets, or payment calls by stubbing the HTTP session:

```bash
PYTHONPATH=01-features/08-agents-that-transact/02-use-cases/pay-for-data/agent python3 - <<'PY'
from catalog import fetch_live_catalog

class DummyResponse:
    encoding = 'utf-8'
    def raise_for_status(self): pass
    def iter_content(self, chunk_size):
        yield b'{"agents": []}'

class DummySession:
    def get(self, *args, **kwargs):
        print('would fetch', args, kwargs)
        return DummyResponse()

try:
    fetch_live_catalog(session=DummySession())
except Exception as exc:
    print(type(exc).__name__)
    print(exc)
PY
```

Result:

```text
RuntimeError
Missing required environment variables: PAYMENT_MANAGER_ARN, PAYMENT_SESSION_ID, PAYMENT_INSTRUMENT_ID...
```

Patch direction: keep catalog loading independent from payment-session config. For example, have `fetch_live_catalog()` read `HEURIST_CATALOG_URL` directly with the current default, or split `get_config()` into a catalog-only config and a runtime/payment config. Separately, either align `MANAGER_ARN` vs `PAYMENT_MANAGER_ARN`, or have the host-side script adapt the stored `MANAGER_ARN` value when invoking runtime/session code.

Additional local check: `python3 -m py_compile` over the new Python files passed.

joshuamarksmith:
Updated, thank you. `fetch_live_catalog()` now accepts an optional `catalog_url` parameter that when provided, tells the system to skip `get_config()` entirely. The host-side script passes the URL directly from the environment, so it no longer requires `PAYMENT_SESSION_ID` or `PAYMENT_MANAGER_ARN` to be set at catalog-sync time.

TateLyman:
Re-checked latest `fcd6d6d`.

The catalog-sync blocker I noted is fixed: `fetch_live_catalog()` now accepts `catalog_url`, skips `get_config()` when it is provided, and the host script passes `HEURIST_CATALOG_URL` directly during Step 5.

Local no-payment verification:

```bash
PYTHONPATH=agent python3 - <<'PY'
from catalog import fetch_live_catalog

class DummyResponse:
    encoding = "utf-8"
    def raise_for_status(self):
        pass
    def iter_content(self, chunk_size):
        yield b'{"agents": []}'

class DummySession:
    def __init__(self):
        self.calls = []
    def get(self, *args, **kwargs):
        self.calls.append((args, kwargs))
        return DummyResponse()

session = DummySession()
payload = fetch_live_catalog(session=session, catalog_url="https://example.test/catalog.json")
print(payload)
print(session.calls)
PY
python3 -m py_compile pay_for_data.py agent/catalog.py agent/config.py agent/main.py agent/sync_registry.py
```

Result: the dummy catalog fetch returns `{"agents": []}` and does not require `PAYMENT_SESSION_ID`, `PAYMENT_MANAGER_ARN`, or `PAYMENT_INSTRUMENT_ID`. `py_compile` also passes on the new Python files.

One small follow-up I noticed but would not block this fix: the host-side default catalog URL is `https://mesh.heurist.xyz/x402/agents?details=true`, while `agent/config.py` defaults to `https://mesh.heurist.xyz/x402/base-sepolia/agents?details=true`. If those are intentionally different between host sync and runtime refresh, ignore this; otherwise aligning them would prevent host/runtime catalog drift.

mvangara10 (approved):

Issue #1562, opened 2026-05-23
01-features - [Bug] 05-authenticate-and-authorize/01-inbound-auth/01-inbound-auth-cognito undeployable


palbiren:
**Bug Description**

  `01-features/05-authenticate-and-authorize/01-inbound-auth/01-inbound-auth-cognito/inbound_auth_runtime.py` is undeployable.

   ## Issues

   - **Build step crashes immediately.** `uv pip install` is invoked with the
     wrong `--python-platform` value (`manylinux2014_aarch64` instead of
     `aarch64-manylinux2014`), the wrong python flag (`--python` instead of
     `--python-version`), and bypasses the package's own `requirements.txt`
     in favor of hardcoded subprocess args.

   - **`create_agent_runtime` ** `codeConfiguration` is passed as a top-level kwarg (must be nested in `agentRuntimeArtifact`); `containerConfiguration` is set redundantly with an S3 URI as `containerUri`; the code source uses the deprecated `sourceCode.s3.{uri, etag=""}` shape instead of canonical `code.s3.{bucket, prefix}`.

   - **IAM scope too narrow for the agent's model.** Once deployed, the agent's first model call returns `AccessDeniedException` because `bedrock:InvokeModel` is scoped only to `foundation-model/*`, but the
  embedded agent uses `global.anthropic.claude-haiku-4-5-20251001-v1:0` (a global inference profile).

   - **Authenticated invocation parameter is invalid.** Test 7 raises `ParamValidationError` on `bearerTokenCredentials`, which is not a parameter of `invoke_agent_runtime`.

   - **Cleanup leaks the Cognito user pool.** `cleanup()` reads `state.get("user_pool_id")` instead of the actual save location nested under `state["cognito_config"]["user_pool_id"]`.

   PR forthcoming.

PR #1565, opened 2026-05-23
fix: restore 01-inbound-auth-cognito sample to deployable state

palbiren:
**Issue number**: #1562

   Restores `01-features/05-authenticate-and-authorize/01-inbound-auth/01-inbound-auth-cognito/inbound_auth_runtime.py` to a deployable state. The sample failed at multiple points and could not be deployed end-to-end on a fresh AWS account. Verified working end-to-end on 2026-05-22 (deploy → no-auth 401 → Cognito-JWT-authenticated invoke → cleanup).

   ### User experience

   **Before:** running `python inbound_auth_runtime.py` on a clean account crashes immediately at Step 3 because `uv pip install` is invoked with the wrong `--python-platform` value and the wrong python flag. If the
   user manually patches past that, Step 4 fails with `ParamValidationError: Unknown parameter "codeConfiguration"` on `create_agent_runtime`. Even patched past Step 4, the agent's first model call returns
   `AccessDeniedException` (IAM doesn't cover the global inference profile the embedded agent uses), and Step 7 crashes with `ParamValidationError: Unknown parameter "bearerTokenCredentials"`.

   **After:** the script runs end-to-end without manual patches. The agent boots, authenticates against Cognito, invokes Bedrock, and returns a result. `--cleanup` removes every resource the script creates (including
   the Cognito user pool, which previously leaked).

   ### Summary of changes

   All in `inbound_auth_runtime.py`:

   **Build step**
   - `--python-platform` value flipped to uv's expected order (`aarch64-manylinux2014` instead of `manylinux2014_aarch64`)
   - `--python` replaced with `--python-version` (different uv flag)
   - Added `--only-binary :all:` to avoid source builds for native deps
   - Switched from hardcoded subprocess args + `--no-deps` + secondary install to a single `uv pip install -r requirements.txt` invocation; deps now track `requirements.txt` instead of drifting

   **Runtime creation**
   - Removed top-level `codeConfiguration` kwarg; `codeConfiguration` must be nested inside `agentRuntimeArtifact` per current API
   - Removed misuse of `containerConfiguration.containerUri = s3_uri` (containerUri is for ECR images)
   - Switched to canonical `code.s3.{bucket, prefix}` shape; redefined `S3_PREFIX` to be the full object key (`{AGENT_NAME}/code.zip`) since AgentCore validates the referenced object is a `.zip`
   - Wrapped `create_agent_runtime` in retry-with-backoff to absorb the IAM role propagation race (4/8/16/32/64s)
   - Bumped post-role-creation sleep from 10s to 15s

   **IAM**
   - Broadened `bedrock:InvokeModel` resource to include `arn:aws:bedrock:*:{ACCOUNT_ID}:inference-profile/*` — required because the embedded agent uses `global.anthropic.claude-haiku-4-5` (cross-region inference
   profile, not foundation-model). Also added `bedrock:Converse` and `bedrock:ConverseStream` actions used by Strands' default streaming flow.

   **Invocation**
   - Replaced `bearerTokenCredentials` (not a valid `invoke_agent_runtime` parameter) with the canonical `Authorization`-header injection via boto3 `before-send` hook; matches the pattern in `03-m2m-3lo/invoke.py`

   **Cleanup**
   - Read `user_pool_id` from `state["cognito_config"]` (where it is actually saved) instead of from the top level of state, so cleanup actually deletes the Cognito user pool

   ## Checklist

   * [ ] Add your name to [CONTRIBUTORS.md](https://github.com/awslabs/amazon-bedrock-agentcore-samples/blob/main/CONTRIBUTORS.md)
   * [ ] Have you documented **Introduction**, **Architecture Diagram**, **Prerequisites**, **Usage**, **Sample Prompts**, and **Clean Up** steps in your example README?
   * [x] Changes are documented

   ## Acknowledgment

   By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project
   license](https://github.com/awslabs/amazon-bedrock-agentcore-samples/blob/main/LICENSE).

### DIFF
--- a/01-features/05-authenticate-and-authorize/01-inbound-auth/01-inbound-auth-cognito/inbound_auth_runtime.py
+++ b/01-features/05-authenticate-and-authorize/01-inbound-auth/01-inbound-auth-cognito/inbound_auth_runtime.py
@@ -32,6 +32,7 @@ import zipfile
 
 import boto3
 from boto3.session import Session
+from botocore.exceptions import ClientError
 
 # ── Configuration ─────────────────────────────────────────────────────────────
 
@@ -47,6 +48,7 @@ session = Session()
 REGION = session.region_name or "us-east-1"
 ACCOUNT_ID = session.client("sts").get_caller_identity()["Account"]
 S3_BUCKET = f"agentcore-code-{ACCOUNT_ID}-{REGION}"
+S3_PREFIX = f"{AGENT_NAME}/code.zip"
 
 print(f"Region:  {REGION}")
 print(f"Account: {ACCOUNT_ID}")
@@ -259,8 +261,13 @@ def create_execution_role() -> str:
                 "Action": [
                     "bedrock:InvokeModel",
                     "bedrock:InvokeModelWithResponseStream",
+                    "bedrock:Converse",
+                    "bedrock:ConverseStream",
                 ],
-                "Resource": "arn:aws:bedrock:*::foundation-model/*",
+                "Resource": [
+                    "arn:aws:bedrock:*::foundation-model/*",
+                    f"arn:aws:bedrock:*:{ACCOUNT_ID}:inference-profile/*",
+                ],
             },
         ],
     }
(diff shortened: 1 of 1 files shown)


01-features/05-authenticate-and-authorize/01-inbound-auth/01-inbound-auth-cognito/inbound_auth_runtime.py
```diff
@@ -280,94 +287,88 @@ def create_execution_role() -> str:
...
 
-        # Upload to S3
+        # Upload to S3.
         s3 = boto3.client("s3", region_name=REGION)
         try:
-            s3.create_bucket(
-                Bucket=S3_BUCKET,
-                **({"CreateBucketConfiguration": {"LocationConstraint": REGION}} if REGION != "us-east-1" else {}),
-            )
+            if REGION == "us-east-1":
```
BharathiSrini, line 356: can we remove this region hard-coding? Let's use a variable populated by the user at the top of the script.

BharathiSrini (approved):

01-features/05-authenticate-and-authorize/01-inbound-auth/01-inbound-auth-cognito/inbound_auth_runtime.py
palbiren: LocationConstraint value can't be us-east-1 and thus this check

PR #1628, opened 2026-06-06
feat(memory): add high-level AgentCore SDK (MemorySessionManager) path to memory tutorials

di-brasil:
## Summary

Each standalone memory tutorial now runs the same flow two ways, so a reader can compare the raw API against the SDK from one file:

```bash
python <tutorial>.py boto3    # the raw AWS API, no SDK (bedrock-agentcore / -control)
python <tutorial>.py sdk      # the AgentCore SDK (MemorySessionManager), the recommended way
```

The `sdk` path uses the high-level **`MemorySessionManager` / `MemorySession`** API. This PR originally added it as a third way to run each tutorial, alongside a low-level `MemoryClient` one. Per review, those two were redundant (same operations, different method names, and both create/delete through `MemoryClient` anyway), so they're consolidated into the single `sdk` path shown above. The `boto3` path is unchanged.

## What this PR adds

**14 tutorials** gain the `sdk` (MemorySessionManager) path:

| Area | Tutorial | Notable detail in the `sdk` version |
|---|---|---|
| Short-term | events-and-sessions | `add_turns` → events; `get_last_k_turns` read-back |
| Short-term | event-metadata-filtering | metadata on turns + filtered read |
| Short-term | actor-session-isolation | asserts one actor can't read another's data |
| Short-term | branching | `fork_conversation` + `list_branches` |
| Short-term | standard-usage | end-to-end STM walkthrough |
| Long-term | standard-usage | end-to-end LTM walkthrough |
| Long-term | built-in-strategies / semantic | `namespaceTemplates`; recall via `search_long_term_memories` |
| Long-term | built-in-strategies / user-preference | structured-preference recall |
| Long-term | built-in-strategies / summary | summary consolidation |
| Long-term | built-in-strategies / episodic | sets required `reflectionConfiguration`; polls for the slower extract→reflect pipeline |
| Long-term | namespaces | hierarchical namespace retrieval |
| Long-term | retrieval | record + citation retrieval |
| Long-term | record-metadata | waits until records are searchable, then filters |
| Long-term | batch-apis | direct batch CRUD with a propagation wait (see note) |

## What the `sdk` code looks like

Same shape in every file. Create the resource with `MemoryClient` (control plane), then drive everything else through a `MemorySession` (data plane):

```python
def run_with_sdk(cleanup: bool = False) -> None:
    from bedrock_agentcore.memory import MemoryClient, MemorySessionManager
    from bedrock_agentcore.memory.constants import ConversationalMessage, MessageRole

    # MemorySessionManager is data-plane only, so the resource is still
    # created/deleted through MemoryClient.
    client = MemoryClient(region_name=REGION)
    memory = client.create_memory_and_wait(
        name=..., strategies=[{"semanticMemoryStrategy": {
            "name": "UserFacts",
            "namespaceTemplates": [NAMESPACE_TEMPLATE],   # current field (not `namespaces`)
        }}], event_expiry_days=30)

    manager = MemorySessionManager(memory_id=memory["id"], region_name=REGION)
    session = manager.create_memory_session(actor_id=ACTOR_ID, session_id=SESSION_ID)

    # add_turns takes ConversationalMessage objects → one create_event
    session.add_turns(messages=[ConversationalMessage(text, MessageRole[role]) for role, text in TURNS])

    hits = session.search_long_term_memories(query=q, namespace=ns, top_k=3)   # see version note
    #   each hit is a dict-like MemoryRecord: hit["content"]["text"], hit.get("score")
```

## Design decisions

- **Current API only.** Uses `namespaceTemplates` (not the deprecated `namespaces`) and `search_long_term_memories(namespace=…/namespace_path=…)` (not the deprecated `namespace_prefix=`).
- **Control vs. data plane made explicit.** `MemorySessionManager` only does data-plane work, so each tutorial creates/deletes the memory with `MemoryClient` and routes events + retrieval through the session. A comment says so, because this trips people up.
- **One turn-builder pattern.** Turns are built with `ConversationalMessage` + `MessageRole` and written via `add_turns`, which maps to a single `create_event`.
- **Episodic polls for the real pipeline.** The episodic tutorial sets `reflectionConfiguration` and polls for extraction *and* reflection (much slower than single-step extraction), returning as soon as records appear instead of waiting a fixed time and risking an empty read.

## Two real fixes that came up while writing this

Both are applied to the affected tutorials:

- **Records aren't readable the instant you write them.** After `BatchCreateMemoryRecords`, an immediate update/list can raise `ResourceNotFoundException` or return 0 because the records aren't indexed yet. The batch-apis and record-metadata tutorials now wait for the records to become readable before updating/filtering, instead of assuming they're available right away.
- **Timestamp type.** Batch-record timestamps now use a real `datetime` (the field is datetime-typed), not a numeric string.

> Note: the companion PR #1629 hardens the batch-records wait further. Live testing showed "the record shows up in `List`" and "the record is updatable" aren't the same moment, so #1629 retries the actual update/delete on `ResourceNotFoundException` rather than polling `List`. If both PRs land, #1629's version is the one to keep for batch-apis.

## How to test

Both paths were **run for real** in `us-west-2` (using `global.anthropic.claude-sonnet-4-6` where a model is needed). Each run creates a memory, demonstrates the pattern, and deletes the memory it created, so nothing is left behind.

```bash
python <tutorial>.py sdk     # e.g. 02-long-term-memory/01-built-in-strategies/semantic.py
```

- ✅ All 14 `sdk` runs work end-to-end and self-clean.
- ✅ Semantic / user-preference / summary recall returns the expected records.
- ✅ Episodic produces episode + reflection records (after the slower pipeline).
- ✅ actor-session-isolation asserts no cross-actor leakage. The check is written to actually fail if isolation broke, not just pass quietly.
- ✅ Branching forks and lists branches correctly.

## Reviewer note (please read before merging)

The `sdk` recall uses `namespace=`, which exists only in **bedrock-agentcore ≥ 1.14**. Older releases accept only the deprecated `namespace_prefix=`. A normal `pip install bedrock-agentcore` pulls the latest (≥1.14), so this works out of the box and was verified on 1.14. The only failure path is running against an older, already-installed SDK. Each tutorial's docstring notes the requirement.

Credentials come only from the standard AWS provider chain. No secrets in the code.


di-brasil:
**Reviewer note — SDK version requirement for the `session` surface** (for Akarsha's call)

The new `session` surface retrieves with `search_long_term_memories(..., namespace=...)`. That `namespace=` / `namespace_path=` argument only exists in **bedrock-agentcore ≥ 1.14** — older releases only accept the (now deprecated) `namespace_prefix=`, so on an older SDK the `session` recall calls would raise `TypeError: unexpected keyword argument 'namespace'`.

In practice this is fine: the tutorials install with an unpinned `pip install bedrock-agentcore`, which resolves to the latest (≥ 1.14) where `namespace=` works, and the whole surface was verified live on 1.14.0. The only way to hit the error is to run against a pre-1.14 SDK that's already installed and not upgraded.

Open question for review — two reasonable options, happy to do either:
1. **Leave as-is** — `namespace=` is the current, correct API; the rest of the repo is also unpinned, so this is consistent.
2. **Add a small safeguard** — either a one-line "requires `bedrock-agentcore>=1.14` for the `session` surface" note in the affected tutorials, or fall back to the portable `namespace_prefix=` on recall.

No change made yet — flagging it so the reviewer can decide.

di-brasil:
@akshseh — flagging the SDK-version note above for your call: keep `namespace=` as-is (needs bedrock-agentcore ≥1.14), add a small version note, or fall back to the portable `namespace_prefix=`. Happy to make whichever change you prefer.

01-features/04-manage-context-of-your-agent/memory/01-short-term-memory/01-events-and-sessions/events-and-sessions.py
```diff
@@ -6,9 +6,10 @@
     - GetEvent fetches one event in full
     - ListSessions discovers prior sessions for an actor
 
-Two surfaces, same flow:
-    python events-and-sessions.py boto3
-    python events-and-sessions.py sdk
+Three surfaces, same flow:
+    python events-and-sessions.py boto3      # raw boto3 (bedrock-agentcore / -control)
+    python events-and-sessions.py sdk        # AgentCore SDK, low-level MemoryClient
```
akshseh: Good callout. lets consolidate the sdk and session to use the sessionManager.

di-brasil: Done in `0c9bbc5`. Consolidated `sdk` and `session` into a single `sdk` option built on `MemorySessionManager`, across all 14 tutorials.

Each tutorial now has two ways to run it:

```
python <tutorial>.py boto3    # the raw AWS API, no SDK
python <tutorial>.py sdk      # the AgentCore SDK (MemorySessionManager), the recommended way
```

What changed: the old low-level `sdk` option (MemoryClient) was redundant with `session` since both ran the same operations under different names and both create/delete through `MemoryClient` anyway. So I dropped the low-level one and renamed the `MemorySessionManager` one to `sdk`. Create and delete still go through `MemoryClient` because the session manager is data-plane only.

Verified: all 14 `sdk` runs work end to end in us-west-2. Recall returns real records, the batch create/update/delete cycle works, and episodic prints its episode and reflection records. One related fix in the same commit: episodic used to wait a fixed ~16 minutes then query once, so if reflection records surfaced just after the wait it printed nothing even though extraction had worked. It now polls and returns as soon as records appear.

Note: the `sdk` run searches with `search_long_term_memories(namespace=...)`, which needs `bedrock-agentcore >= 1.14`. Older versions only accept the deprecated `namespace_prefix=`. Each tutorial's docstring now says so.

akshseh (dismissed):

akshseh (approved):

PR #1629, opened 2026-06-07
fix(memory): correctness & hygiene fixes across the AgentCore Memory tutorials

di-brasil:
## What this PR does

This is a clean-up pass over the **AgentCore Memory tutorials** (`01-features/04-manage-context-of-your-agent/memory/`). While going through them I found a number of places where the tutorial code either crashes when you run it, silently does the wrong thing, or tells you something about the SDK that isn't true. This PR fixes those.

I grouped the work into three commits so it's easy to review one layer at a time:

- **`4db8cf9` — things that break (P0):** code that throws an error or behaves wrong when you run the tutorial as written.
- **`433cc18` — recall & lifecycle (P1):** memory that "works" but quietly returns nothing, plus a few fragile setup patterns.
- **`4ea4466` — tidy-up (P2):** a swallowed bug, a garbled prompt, a duplicate dependency, and two stray internal files.

Every behavioral fix was **run for real** against AWS (with the actual framework — Strands / LlamaIndex / LangGraph — and the SDK version the tutorial actually installs), not just checked that it compiles. Test memories were cleaned up after each run.

---

## Commit 1 — `4db8cf9`: things that break when you run them

- **events-and-sessions:** the boto3 `create_event` call was missing the required `eventTimestamp`, so it crashed immediately. Added it.
- **branching:** the code passed `include_parent_events=…`, but the real argument is `include_parent_branches=…`, so it raised a `TypeError`. Fixed the name.
- **record-streaming:** `streamDeliveryResources` was written as a flat list, but the API expects a nested structure (`resources → kinesis → contentConfigurations`). Fixed in both the boto3 and SDK versions and in the README.
- **batch-apis:** right after you create records in bulk, trying to update or delete them often fails for a while because the records aren't readable yet (they take a few seconds to tens of seconds to show up). The tutorial assumed they were available instantly. Added a small retry that waits for them to land.
- **episodic:** this one was the most broken. It was missing the required `reflectionConfiguration` (so creating the memory failed outright), and it waited a blind 20 minutes with no feedback. Now it: adds the required config; waits by *polling* and returns as soon as records appear; stores episodes per-session and reflections per-actor (and reads them back together); includes a tool-result turn so the episode has richer content; and ends each conversation with a closing turn so the service actually recognizes the episode as finished. Verified end to end: both example episodes now extract, with their cross-episode reflections.
- **runtime & security integrations:** the code indexed the invoke response like `response["response"]`, but the response is a `requests.Response` object — that throws. Switched to `.json()` and removed a dead code block.
- **memory-browser backend:** the local server bound to `0.0.0.0`, which exposes an AWS-credentialed memory proxy to your whole network. Now it binds to `127.0.0.1` by default (and you can override host/port with env vars if you really want to).

## Commit 2 — `433cc18`: memory that looks fine but quietly returns nothing

- **LlamaIndex memory-tool examples (4 tutorials):** the agent *saved* memories to one namespace but *searched* a different one (`/strategies/`), so it always got back nothing. Now it reads from the same namespace it writes to. (I kept the existing `namespace_prefix=` argument on purpose — the newer `namespace=`/`namespace_path=` arguments don't exist in the older SDK version these tutorials install, so switching would break them.)
- **debugging-agent:** the retrieval tools used a hard-coded search string instead of the user's actual question, referenced a `session_id` that didn't exist in that scope, and had a `count_tokens()` helper that actually counted characters. Fixed all three.
- **customer-support (×2), meeting-notes, healthcare:** these used a brittle "try to create, catch the error, then list everything and match by name" pattern to get-or-create a memory. Replaced it with the SDK's built-in `create_or_get_memory`, which does exactly that, correctly.
- **healthcare, personal-agent:** the error handler referenced `memory_id` before it was guaranteed to be set, so an early failure would raise a confusing `NameError` instead of the real error. Initialized it to `None` first.
- **structured-metadata:** a comment claimed the SDK doesn't expose `indexed_keys` and reached into an internal client to work around it. The SDK *does* expose it now — switched to the normal call and removed the stale comment.
- **nutrition (LangGraph):** when the IAM execution role already existed, the tutorial updated its trust policy but didn't wait for that change to propagate before creating the memory — so a re-run failed with "please provide a role with a valid trust policy." Added the same short wait the create path already had. (Found while running it twice; confirmed the re-run now succeeds.)
- **Honesty notes (no behavior change):** four tutorials named/described as "episodic" are actually configured with a semantic or custom strategy. Rather than silently rewrite them, I added a short note explaining what they really use and pointing to the real episodic example.
- **guardrails README:** it claimed the example filters content going into and out of memory. It doesn't — it only filters the model's output. Reworded to match what the code actually does, with a note on what real memory-side filtering would require.
- **LlamaIndex README:** the default model id was a retired model; updated it to match the model the scripts actually use.

## Commit 3 — `4ea4466`: small fixes and tidy-up

- **travel-planning-branching:** the conversation-history loader read `message.content.get("role")` and `…get("content").get("text")` — one level too deep — so the role was always "unknown" and the text always empty, and the loaded history was silently thrown away. Fixed the access; verified live (it went from 0 messages recovered to the full history).
- **travel-planning (×2):** a copy-paste had mangled a prompt into "direct flimid-range, city center, pool, standard room" (flight and hotel terms fused together). Rewrote it as a clean flight-only request that matches the surrounding turn.
- **with-strands requirements:** listed both `duckduckgo-search` and `ddgs`, but the code only imports `ddgs`. Dropped the unused one (confirmed the trimmed file still installs and imports).
- **stray files:** removed `notes.md` and `documentation.md` from the memory root — internal authoring/scratch notes that shouldn't ship in the public tutorials (nothing in the repo links to them).

---

## What I deliberately did **not** change

A few flagged items turned out to be non-issues or risky, so I left them alone (happy to revisit any):

- A couple of "bugs" were actually correct code (e.g. the memory-browser SDK calls all exist; the LangGraph store legitimately uses tuple namespaces).
- I did **not** pin every `requirements.txt` to a newer SDK — doing so actually breaks dependency resolution for the LlamaIndex tutorials and would conflict with the portability fix above.
- A handful of larger hygiene items (model-id prefix sweep, IAM scope-tightening, replacing remaining blind sleeps, de-duplicating boilerplate, doc-link fixes) are intentionally **not** in this PR to keep it focused on correctness. They can follow in a separate change.

## Testing

- Behavioral fixes were run end-to-end against AWS in `us-west-2`: episodic extraction, batch retry, recall (with a before/after check proving it returned 0 records before and real records after), the debugging-agent, meeting-notes, travel-planning, and nutrition tutorials.
- Dependency and file-deletion changes were verified by actually installing the trimmed requirements and scanning the whole repo for references.
- All changed Python compiles; no secrets introduced; test resources cleaned up after each run.


di-brasil:
@akshseh — this is the Section-1 correctness/hygiene PR (fixes across the memory tutorials), companion to #1628. Three commits (P0 → P1 → P2), each behavioral fix run end-to-end. Would appreciate your review when you have a chance.

akshseh (approved):

akshseh (approved):

PR #1671, opened 2026-06-13
fix(payments): pin agentcore CLI to 0.19+ for T02 deploy

fahadfa-aws:
## Issue

The Tutorial 02 prereqs say `npm install -g @aws/agentcore` with no version pin. The CLI's 0.14 release scaffolds `<project>/agentcore/aws-targets.json` as an empty array, so `agentcore deploy` exits with \`Target \"default\" not found in aws-targets.json\` before the CDK synth ever runs. A fresh user following the README hits this and the error gives no clue that the CLI version is the cause. Reproduced on 2026-06-12.

## Changes

- Prereqs and CLI Commands block: pin to `@aws/agentcore@^0.19.0`.
- New troubleshooting entry: \"agentcore deploy fails with `Target \"default\" not found in aws-targets.json`\" with the upgrade path and a hand-written `aws-targets.json` fallback for users who can't upgrade.

Docs only. No code change.

## Verification

- 0.14 reproduced the failure on 2026-06-12 against `us-west-2`.
- 0.19.0 deploys cleanly using the same script and account.
- Re-read of the cited file:line at upstream HEAD `3a8d5352`:
  - `deploy_payment_agent.py:154` is the `subprocess.run([\"agentcore\", \"deploy\", \"-y\"], ...)` call.
  - No version probe exists today; this PR is README-only.

### DIFF
--- a/01-features/08-agents-that-transact/00-getting-started/02-deploy-to-agentcore-runtime/README.md
+++ b/01-features/08-agents-that-transact/00-getting-started/02-deploy-to-agentcore-runtime/README.md
@@ -49,15 +49,15 @@ App Backend                          AgentCore runtime
 - Tutorial 01 completed (understand the local agent + plugin flow)
 - Wallet funded with testnet USDC from [faucet.circle.com](https://faucet.circle.com/)
 - Python 3.10+
-- Node.js 20+ and the AgentCore CLI: `npm install -g @aws/agentcore`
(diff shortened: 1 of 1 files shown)


fahadfa-aws:
Closing — the version pin (`^0.19.0`) is a 1–4 week bandage that goes stale the moment `@aws/agentcore@0.20.0` ships. Reopening with a more durable approach: detect the empty `aws-targets.json` at deploy time so the script self-heals across CLI versions instead of pinning to a known-good version that a future release will obsolete.

PR #1748, opened 2026-06-26
Add Receipts IDP agent (Intelligent Document Processing) with a model degradation ladder

di-brasil:
### What this adds

An event-driven **Intelligent Document Processing** sample on Amazon Bedrock AgentCore. A receipt uploaded to S3 is processed by a **dual-agent** pipeline (extractor → independent validator), persisted per-user through an AgentCore **Gateway** with a **Cedar** spend guardrail, and fully traced + evaluated in CloudWatch. A user can also **chat** with a read-only assistant about their own expenses.

Its distinct contribution over the existing claims sample is a **model degradation ladder**: when Amazon Bedrock returns `503` (capacity), the pipeline steps to a different model (global inference profiles, Opus 4.8 → 4.7 → 4.6 → Sonnet 4.6 → defer-to-SQS) and recovers automatically — without a redeploy — so a model outage degrades quality safely instead of dropping work.

### Exercises six AgentCore services

Runtime (the three agents), Gateway (governed MCP tool surface), Policy/Cedar (the spend guardrail), Observability (OTel → CloudWatch, spans tagged with the degradation rung), Evaluations (LLM-as-judge + online metrics), and Memory (per-user facts/summaries).

### Highlights

- **Dual-agent extraction** — an independent validator owns the auto-file-vs-review decision (avoids confirmation bias).
- **Model degradation ladder** — config-driven model on AWS AppConfig; reactive in-agent `503` step-down + a proactive account-level control loop (CloudWatch alarm → controller Lambda → step the active rung); L4 buffers to SQS with a jittered drain (no thundering herd).
- **Deterministic guardrail** — Cedar gates the write on the tool input (amount ≥ $2,000 → review), independent of the agents.
- **Conversational query mode** — a read-only assistant answers natural-language questions about the user's own expenses; identity is a KMS-HMAC-signed token (not the request body) and the read tools are pinned server-side, so a user cannot read another user's data (IDOR designed out, with a negative test).
- **Operational audit** — a per-receipt ProcessingRuns ledger: one durable, queryable fate per receipt (incl. errors), with push alerts on errors.
- **One-command deploy + clean teardown**; the agent depends only on env/AppConfig (a replaceable deploy seam).

### Testing

No mocks — every end-to-end test runs against real AWS. Unit tests for pure logic; live e2e for the full pipeline, the front door, Cedar, the ladder + control loop + L4 drain, the run ledger, and the conversational path including an **IDOR negative test**. A self-verifying `scripts/user-demo.sh` walks the whole user journey. Latest full clean-deploy run: 19 passed, 1 skipped (the live-503 sim, unit-tested instead), 0 failures.

### Docs

`README.md`, `docs/ARCHITECTURE.md` (with diagrams), `docs/CONFIGURATION.md`, `docs/tutorial.md`, `docs/deployment.md`, and 16 ADRs under `docs/decisions/`.

### Notes for reviewers

- Experimental/educational sample (matches the repo's standard disclaimer).
- Requires four `global.` Bedrock inference profiles enabled (the ladder rungs).
- Known boundaries (intentional, documented): the live `503` step-down is unit-tested (a real 503 can't be triggered on demand); the human *review action* (approve/correct) is out of scope — the sample routes to review and records it.


02-use-cases/02-workflow-automation-agents/receipts-intelligent-document-processing-agent/destroy.sh
```diff
@@ -0,0 +1,31 @@
...
+REGION="${1:-us-west-2}"
+STACK="${RECEIPTS_STACK:-AgentCore-ReceiptsAgent-dev}"
+export AWS_REGION="$REGION"
+export AWS_DEFAULT_REGION="$REGION"
+
+echo "🧹 Destroying $STACK in $REGION..."
+
+if ! aws cloudformation describe-stacks --stack-name "$STACK" --region "$REGION" >/dev/null 2>&1; then
+  echo "✅ Stack $STACK not present — nothing to destroy."
+  exit 0
```
wilmatos, line 33: We may want to handle DELETE_FAILED or orphaned AgentCore control-plane resources. AgentCore can hit these scenarios due to control-plane resource ordering.  At minimum, add a retry loop or point to user guidance for cleanup.

02-use-cases/02-workflow-automation-agents/receipts-intelligent-document-processing-agent/.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
+name: receipts-idp-ci
```
wilmatos, line 1: Can you address CKV2_GHA_1: Top-level permissions should be scoped?  
Add `permissions: read-all` at the workflow level and override per-job where write is needed.

Will need to add a `permissions: { contents: read, id-token: write }` on the `e2e` job.

02-use-cases/02-workflow-automation-agents/receipts-intelligent-document-processing-agent/destroy.sh
di-brasil: Good catch, thank you. Fixed in c31d142.

`destroy.sh` now recovers from `DELETE_FAILED` automatically instead of exiting on an opaque `cloudformation wait` error:

1. Polls `delete-stack` to a terminal status.
2. On `DELETE_FAILED`, retries the plain delete once — most control-plane ordering orphans are transient and clear on a second pass.
3. If specific resources are still stuck, re-issues the delete with `--retain-resources <LogicalId ...>` (valid only for a stack already in `DELETE_FAILED`) so the stack and everything billable is removed rather than left wedged.
4. Prints each retained resource as `ResourceType → PhysicalResourceId`, and points to a new **Teardown & DELETE_FAILED recovery** section in `docs/deployment.md` with the exact `bedrock-agentcore-control delete-gateway` / `delete-gateway-target` / `delete-agent-runtime` calls for the manual cleanup.

All four teardown paths (clean, stack-absent, recover-via-retry+retain, terminal-failure) were verified against a mocked AWS CLI.

02-use-cases/02-workflow-automation-agents/receipts-intelligent-document-processing-agent/.github/workflows/ci.yml
di-brasil: Done in 155b3a2. Added `permissions: read-all` at the workflow level and scoped each job to least-privilege:

- `unit`: `contents: read` (checkout only)
- `e2e`: `contents: read` + `id-token: write` (required for the OIDC role assumption via `aws-actions/configure-aws-credentials`)

Verified locally with checkov 3.3.6: `CKV2_GHA_1` goes from FAILED to **PASSED**, and no new finding is introduced. I also audited every step that consumes a `GITHUB_TOKEN` scope to confirm none lost a permission it relies on (`make e2e` runs on the assumed AWS role, not the GitHub token).

One transparency note: checkov also reports a pre-existing `CKV_GHA_7` on this file ("workflow_dispatch inputs must be empty"), triggered by the `run_e2e` toggle that lets a maintainer launch the live deploy→assert→destroy e2e on demand. It predates this change and the repo's ASH scan treats it as advisory (`continue-on-error: true`). Happy to drop the input if you'd prefer a clean checkov pass — that just means the e2e would run on `main` only, losing the manual trigger. Let me know your preference.

wilmatos: Verified. Looks good now.

wilmatos: I'm ok with the CKV_GHA_7 if it doesn't block merging.

wilmatos (approved):
Verfied. Looks good to me.

akshseh (approved):

PR #1849, opened 2026-07-29
fix(memory): wait for eventual consistency of directly-written records, add missing skip-extraction sdk surface

di-brasil:
Three samples report a result the run does not support. All three exit 0 while doing it, so nothing flags the failure. Found by running every command in these folders' READMEs against a live account and checking the retrieved records rather than the exit code.

Each commit is one root cause and can be reviewed on its own.

## `b75e005c` — reading directly-written records before they propagate

Records written with `BatchCreateMemoryRecords` are eventually consistent, and both samples that use them read back too early.

**`07-batch-apis/batch-create-update-delete.py`** creates 3 records, deletes 1, then prints `Remaining (0)` when 2 should remain. Reproduces on both surfaces.

The file already retries `ResourceNotFoundException` for `BatchUpdate`/`BatchDelete` via `_retry_until_propagated`. That helper cannot cover the final read, because an unpropagated `ListMemoryRecords` does not raise — it succeeds and returns an empty page. `List` also lags the furthest behind: measured live, update and delete both succeeded on the first attempt while the surviving records took ~87s to become listable. The module docstring said the opposite (that `List` leads updatability) and is corrected.

**`06-record-metadata/structured-metadata.py`** filters on `region=EU` after a fixed `sleep(35)`, with a comment putting propagation at ~30s. Measured over three runs, records became searchable at 75s, 86s and 97s. The sleep was short every time; the runs that passed did so only under load that slowed them into the window.

Both now poll for the **expected record count** rather than for a non-empty result. That matters: an earlier version of the metadata fix broke on the first hit and returned 1 of the 2 EU records, which reads as though the filter dropped one. The counts are deterministic here because the records are written directly rather than extracted.

## `4eac09a8` — `skip-extraction.py sdk` silently ran the boto3 path

`08-manage-extraction/README.md` documents both `boto3` and `sdk`, but the script defined only `run_with_boto3` and its entrypoint read nothing except `--cleanup`. Passing `sdk` ran boto3. It exits 0 with correct records, so the only tells were `[boto3]` labels in an `sdk` run.

This is the only such mismatch in the area — all 16 README-documented `sdk` commands under `00-getting-started`, `01-short-term-memory` and `02-long-term-memory` were checked against the functions their scripts define.

Adds the missing surface rather than removing the documented command, keeping the two-way structure every other sample has. It uses `MemoryClient.create_event(extraction_mode=...)` rather than the session API, because `add_turns()` does not accept `extraction_mode` and choosing extraction per event is the whole lesson. `main()` now dispatches like its siblings, so an unknown surface exits 1 instead of silently running boto3.

## Verification

All live in `us-east-1`, `boto3 1.43.58` / `bedrock-agentcore 1.19.0`, run in parallel.

| Sample | Before | After |
|---|---|---|
| `batch-create-update-delete.py` (both surfaces) | `Remaining (0)` | `Remaining (2)`, update reflected, deleted record absent |
| `structured-metadata.py` (both surfaces) | `EU-only results (0)` | `EU-only results (2)`, no US record through the filter |
| `skip-extraction.py sdk` | ran the boto3 path | runs the SDK path; 12 events stored, 0 skipped events extracted |

4 runs each for the first two (2 trials x 2 surfaces), all correct. Unknown-surface dispatch confirmed to exit 1 without an AWS call. No leaked memory resources.

Passes `ruff check` and `ruff format --check` per `.github/workflows/python-lint.yml`.

## Relationship to #1847 and #1848

No file overlap with either, so this merges independently in any order. Same theme as #1848 (an async operation treated as synchronous) but a different mechanism: #1848 waits on model-driven *extraction*, this waits on *propagation* of records written directly. Note `08-manage-extraction/README.md` is touched by #1847, not here — the `sdk` command this commit implements was already documented there.

### DIFF
--- a/01-features/04-manage-context-of-your-agent/memory/02-long-term-memory/06-record-metadata/structured-metadata.py
+++ b/01-features/04-manage-context-of-your-agent/memory/02-long-term-memory/06-record-metadata/structured-metadata.py
@@ -34,6 +34,30 @@ import uuid
 REGION = os.getenv("AWS_REGION", "us-east-1")
 ACTOR_ID = "tenant-acme"
 NAMESPACE = f"/tenants/{ACTOR_ID}/notes/"
+SEARCHABLE_WAIT_SECONDS = 180  # polling budget; records became searchable at 75-97s across runs
+EXPECTED_EU_RECORDS = 2  # of the 3 records written below, 2 carry region=EU
+
+
+def _search_until_visible(
+    op, expected: int = EXPECTED_EU_RECORDS, *, max_wait: int = SEARCHABLE_WAIT_SECONDS, poll: int = 10
+) -> list:
+    """Call `op()` until it returns `expected` records, or the deadline passes.
+
+    Directly-written records are eventually consistent. Measured against a live
+    account over three runs they became searchable at 75s, 86s and 97s after
+    BatchCreate, so any single fixed sleep is a guess that eventually loses.
+
+    This waits for the expected COUNT, not merely for a non-empty result: the index
+    populates record by record, so a run that stops at the first hit can report 1 of
+    the 2 EU records and look like the filter dropped one. The count is deterministic
+    here because the records are written directly rather than extracted.
+    """
+    deadline = time.time() + max_wait
+    while True:
+        hits = op()
+        if len(hits) >= expected or time.time() >= deadline:
+            return hits
+        time.sleep(poll)
 
 
 def _records() -> list[dict]:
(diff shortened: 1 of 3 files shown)


01-features/04-manage-context-of-your-agent/memory/02-long-term-memory/06-record-metadata/structured-metadata.py
```diff
@@ -34,6 +34,30 @@
 REGION = os.getenv("AWS_REGION", "us-east-1")
 ACTOR_ID = "tenant-acme"
 NAMESPACE = f"/tenants/{ACTOR_ID}/notes/"
+SEARCHABLE_WAIT_SECONDS = 180  # polling budget; records became searchable at 75-97s across runs
```
akshseh: Same point here. Make it 90s.

di-brasil: Changing to 100 to match the value from PR #1848 and also because, in most tests, the time exceeded 90.

di-brasil:
Rebased onto `main` to clear the conflict. It was a rename, not a content clash: #1980 moved `07-batch-apis/` to `07-skip-STM/` and carried `batch-create-update-delete.py` over unchanged, so GitHub flagged modify-vs-delete. The fix now applies at the new path. No lines from #1980 were touched.

Also dropped `PROPAGATION_WAIT_SECONDS` from 180s to 120s. Records became listable ~87s after the delete in testing, so 120s keeps a margin without the excess. This matches the tighter budgets @akshseh asked for in #1848.

One thing worth a look while this folder is fresh. The new README added in #1980 says:

> Created records are eventually consistent. A record reported `SUCCEEDED` is not immediately updatable or listable; `BatchUpdate`/`Delete`/`List` can raise `ResourceNotFoundException` for seconds to a minute. Retry the dependent operation (the script shows the pattern) instead of assuming availability.

Two small mismatches with what I measured, neither touched in this PR:

1. **"the script shows the pattern"** was not true for the final `List`. `_retry_until_propagated` only retries `ResourceNotFoundException`, and an unpropagated `ListMemoryRecords` raises nothing at all, it just returns an empty list. That is why the script printed `Remaining (0)`. This PR adds `_list_until_visible`, which polls until the expected count appears, so the sentence holds now.
2. **"seconds to a minute"** looks slightly low. Records took ~87s to become listable after the delete across runs here. Happy to send a one-line follow-up if you want the README to say "seconds to over a minute", but I left it alone rather than edit a README from another PR.

akshseh (approved):

PR #1857, opened 2026-07-30
fix(01-harness): make getting-started work on a fresh AWS account

rmncardoso:
**Issue number**: _to follow — tracking issue being opened_

```
Changes to 01-features/01-harness/00-getting-started and the shared utils/ helpers
it imports, because the getting-started sample fails on a fresh AWS account and
leaves the account in a state the other samples can't recover from.
```

### Scope

`00-getting-started`, the shared `utils/` helpers it imports (`client.py`,
`iam.py`, new `harness.py`) and `requirements.txt`. The other 18 samples in this
folder share several of these defects and follow in separate PRs to keep this one
reviewable.

### User experience

Before — `python getting_started.py` on a clean account:

```
  [1] CREATING
  ... [12] CREATING
=== Step 3: Invoke Agent (Claude Haiku) ===
botocore.errorfactory.ValidationException: Harness is not READY
```

…then exits without deleting the IAM role it created. Every sample here shares the
role name `HarnessExecutionRole`, and `create_harness_role()` returned early when
the role already existed — so that stale role breaks the next sample too, with
nothing pointing at the cause.

After — the sample completes. If creation genuinely fails it raises immediately
with the service's `failureReason`, and cleanup runs either way.

### What changed

- **`getting_started.py`** — the wait-for-READY loop was `for i in range(12)` with
  a 5s sleep: a hard 60s ceiling on an operation measured at ~150s, and it fell
  through instead of raising. Now uses the shared poller. Harness creation moved
  inside the `try` so a `CreateHarness` failure can't leak the role created just
  above it.
- **`utils/harness.py`** (new) — one correct `poll_harness_status()`. This loop
  appears 16 times across the folder in 4 incompatible variants, most checking for
  a `FAILED` status that isn't in the enum (`CREATE_FAILED` / `UPDATE_FAILED` /
  `DELETE_FAILED`), so they can't detect failure at all.
- **`utils/iam.py`** — least privilege (scoped Bedrock and Logs resources, 18
  enumerated AgentCore actions in place of `*Memory*`-style wildcards that
  silently included `Delete*`, `aws:SourceAccount` on the trust policy) and full
  idempotency, so re-running repairs an existing role. Deletion now removes every
  attached policy, which a leftover would otherwise block with
  `DeleteConflictException`.
- **`utils/client.py`** — honour `AWS_REGION` as well as `AWS_DEFAULT_REGION` (an
  empty or missing var produced `Invalid endpoint:
  https://bedrock-agentcore-control..amazonaws.com`); default data-plane
  `read_timeout=900`, since botocore's 60s default is shorter than one agent turn
  and cut the stream off mid-response.
- **`README.md`** — "Wait up to 60 seconds" contradicted the ~150s reality; also
  documents `iam:UpdateAssumeRolePolicy`.
- **`requirements.txt`** — `boto3>=1.43.32`, the first release whose
  `bedrock-agentcore` model knows `awsSkills`; unpinned, it resolves to whatever
  is installed and older versions raise `ParamValidationError`.

### DIFF
--- a/01-features/01-harness/00-getting-started/README.md
+++ b/01-features/01-harness/00-getting-started/README.md
@@ -71,10 +71,19 @@ Your code
 ## Troubleshooting
 
 ### Issue: `CREATING` status never transitions to `READY`
-**Solution**: Wait up to 60 seconds. The first `create_harness` takes longer as the control plane provisions resources. The polling loop in the script handles this.
+**Solution**: Give it a few minutes. A new harness has been observed taking around
+150 seconds to reach `READY` on the public network, and longer in VPC network mode
+or when a custom container image has to be pulled — so a `CREATING` status well
+past the one-minute mark is normal, not a failure. The polling loop in the script
+waits up to 10 minutes and prints the status on every poll.
+
+If it genuinely does fail, the harness moves to `CREATE_FAILED` rather than
+staying in `CREATING`, and the script raises straight away with the service's
+`failureReason` — that message is what tells you the cause (most often a missing
+permission on the execution role).
 
 ### Issue: `NoSuchEntityException` when creating IAM role
-**Solution**: Your AWS credentials may lack IAM permissions. The `create_harness_role()` helper requires `iam:CreateRole`, `iam:PutRolePolicy`. Check your credentials' IAM policy.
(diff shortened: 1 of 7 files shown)


evandrofranco (approved):
ok

Issue #1862, opened 2026-07-30
[Bug] 02-gateway-integration: harness poll times out at 120s and cleanup leaves the harness behind


thor4:
### Sample

`01-features/01-harness/01-advanced-examples/02-gateway-integration`

### What happens

`python gateway_integration.py` fails on a first run in a clean account, and leaves the harness behind when it does:

```
  Harness status: CREATING
  ... (repeats)
TimeoutError: Harness not READY after 120s

Cleaning up...
  Warning: An error occurred (ConflictException) when calling the DeleteHarness operation
  Deleted target: ...
  Deleted gateway: ...
```

The gateway and target are cleaned up. The harness is not, and nothing in the output says how to deal with it.

### Root cause

This file defines its own private copy of `poll_harness_status` instead of using the shared helper added in #1857. That copy carries two defects:

1. **`HARNESS_POLL_TIMEOUT = 120`.** #1857 documented harness provisioning at ~150s on the public network and ~255s in VPC mode, which is why the shared poller allows 600s. A 120s ceiling means the sample gives up before the harness has had time to come up.

2. **It checks for a `FAILED` status that does not exist.** The status enum is `CREATING`, `CREATE_FAILED`, `UPDATING`, `UPDATE_FAILED`, `READY`, `DELETING`, `DELETE_FAILED` — there is no plain `FAILED`. So a harness that genuinely fails to create is treated as "not ready yet", polled until the timeout, and the service's `failureReason` is never surfaced. That is the same defect #1857 called out and fixed centrally.

Cleanup has a third, independent defect. `_cleanup` calls `delete_harness` exactly once, but a harness cannot be deleted while it is still `CREATING` — and the `finally` block runs at precisely the moment that is most likely to be true, right after a step failed mid-provisioning. The call is rejected with `ConflictException`, swallowed by a broad `except Exception`, and the harness is left behind counting against the account's harness quota.

The README compounds it: the troubleshooting entry says to "increase `HARNESS_POLL_TIMEOUT`", which sends the reader to edit a per-sample constant rather than pointing at the shared helper.

### Proposed fix

Delete the private poller and import `utils.harness.poll_harness_status`, which fixes both polling defects at once and stops the constant drifting from the shared value again. Make cleanup wait out `ConflictException` on the same timeout budget, keyed off the error code rather than the exception text. Update the README troubleshooting entry to match.

[...]

PR #1863, opened 2026-07-30
fix(01-harness): inherit the shared harness poller in gateway-integration

thor4:
# Amazon Bedrock AgentCore Samples Pull Request

**Issue number**: #1862

```
Changes to 01-features/01-harness/01-advanced-examples/02-gateway-integration,
because the sample fails on a first run and leaves the harness behind when it
does. It carries a private copy of poll_harness_status instead of using the
shared helper added in #1857, and its cleanup cannot delete a harness that is
still CREATING.
```

### User experience

**Before** — `python gateway_integration.py` in a clean account:

```
  Harness status: CREATING
  ... (repeats)
TimeoutError: Harness not READY after 120s

Cleaning up...
  Warning: An error occurred (ConflictException) when calling the DeleteHarness operation
  Deleted target: ...
  Deleted gateway: ...
```

The gateway and target are cleaned up. The harness is not, and nothing in the output says what to do about it.

**After** — the sample completes. If the harness genuinely fails to create it raises immediately with the service's `failureReason`, and cleanup removes the harness either way, waiting for provisioning to finish first if it has to:

```
  Harness still provisioning, retrying delete in 5s...
  Deleted harness: ...
```

### What changed

- **Deleted the private `poll_harness_status`** and imported `utils.harness.poll_harness_status`. This fixes two defects at once. `HARNESS_POLL_TIMEOUT` was `120`, against provisioning that #1857 measured at ~150s on the public network and ~255s in VPC mode — which is why the shared poller allows 600s. The local copy also checked for a `FAILED` status that **is not in the enum** (`CREATING`, `CREATE_FAILED`, `UPDATING`, `UPDATE_FAILED`, `READY`, `DELETING`, `DELETE_FAILED`), so a harness that failed to create was treated as merely slow, polled to the timeout, and its `failureReason` never surfaced. Importing the shared helper also stops the constant drifting from the shared value again.

- **`_delete_harness()` waits out `ConflictException`.** A harness cannot be deleted while it is still `CREATING`, and `finally` runs at exactly the moment that is most likely to be true — right after a step failed mid-provisioning. The single `delete_harness` call was rejected, swallowed by a broad `except Exception`, and the harness was left behind counting against the account's harness quota with nothing pointing at why. The retry uses the same 600s budget as the poller, keys off `e.response["Error"]["Code"]` rather than matching the exception text, and returns immediately on any other error code.

- **README troubleshooting.** The entry told readers to "increase `HARNESS_POLL_TIMEOUT`", a constant this file no longer defines. Replaced with the real timeout, the `CREATE_FAILED` / `failureReason` check to run when it is exceeded, and a note that the retry-delete log line is expected rather than an error.

### Notes for reviewers

- **Deliberately not included: reuse of an existing harness.** I originally had a `list_harnesses()` "get or create" step, because stuck harnesses accumulate across failed runs. I dropped it. It treats the symptom rather than the cause, and with the poller and cleanup both correct they no longer accumulate. It would also introduce a concurrency hazard the current sample does not have — two runs in one account, the second adopts the first's harness and then deletes it in cleanup while the first is still invoking it — and it obscures what this sample is here to demonstrate, which is the create → use → delete lifecycle. If reuse is wanted, an explicit `--harness-id` alongside `--skip-cleanup` seems the safer shape and I am happy to add it in a follow-up.

- **Scope kept narrow.** Only this sample and its README. The two remaining `BLE001` findings on the file are pre-existing and in code this change does not touch, so I left them per the contributing guidance on not reformatting unrelated code. `ruff` findings on the file go from 4 to 3.

- **Related to #1857**, which fixed the same class of defect in `00-getting-started` and noted the other samples in this folder share it. I checked the open PRs and did not find one covering this sample.

### How it was tested

- Ran the sample end to end against a real account in `us-west-2`, which is how the timeout and the `ConflictException` cleanup failure were found.
- Verified the module imports and resolves `poll_harness_status` from `utils.harness` with the 600s timeout.
- Exercised `_delete_harness` against stubbed clients over five paths: deletes first try; retries through repeated `ConflictException` then succeeds; does **not** retry on a non-conflict error code; gives up at the timeout without raising, so it cannot mask the original failure; and warns without raising when the harness is already gone.
- `ruff check` on the changed file: 3 findings, all pre-existing, none introduced.

## Checklist

* [x] Add your name to [CONTRIBUTORS.md](https://github.com/awslabs/amazon-bedrock-agentcore-samples/blob/main/CONTRIBUTORS.md)
* [x] Have you documented **Introduction**, **Architecture Diagram**, **Prerequisites**, **Usage**, **Sample Prompts**, and **Clean Up** steps in your example README?
* [x] Changes are documented

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/amazon-bedrock-agentcore-samples/blob/main/LICENSE).

### DIFF
--- a/01-features/01-harness/01-advanced-examples/02-gateway-integration/README.md
+++ b/01-features/01-harness/01-advanced-examples/02-gateway-integration/README.md
@@ -73,8 +73,11 @@ changing your agent code.
 ### Issue: Target `FAILED` status
 **Solution**: Check the MCP endpoint is reachable. The default Exa endpoint (`https://mcp.exa.ai/mcp`) requires public internet access from the gateway.
 
-### Issue: `HARNESS_POLL_TIMEOUT` exceeded
-**Solution**: Increase `HARNESS_POLL_TIMEOUT` constant or retry after a few minutes.
+### Issue: `Harness not READY after 600s`
+**Solution**: Harness provisioning is measured at ~150s on the public network and ~255s in VPC mode, so the shared poller in `utils/harness.py` allows 600s. Exceeding that usually means the harness is genuinely stuck rather than slow — check `get_harness` for a `CREATE_FAILED` status and its `failureReason`.
+
+### Issue: cleanup logs `Harness still provisioning, retrying delete`
+**Solution**: Expected. A harness cannot be deleted while it is still `CREATING`, so cleanup waits for provisioning to finish before deleting. It gives up after the same 600s and warns rather than failing the run.
 
 ## AgentCore CLI
 
--- a/01-features/01-harness/01-advanced-examples/02-gateway-integration/gateway_integration.py
+++ b/01-features/01-harness/01-advanced-examples/02-gateway-integration/gateway_integration.py
@@ -47,6 +47,7 @@ sys.path.insert(0, str(Path(__file__).parent.parent.parent))
 
 from utils.iam import create_harness_role
 from utils.client import get_agentcore_client, get_agentcore_control_client
+from utils.harness import HARNESS_POLL_INTERVAL, HARNESS_POLL_TIMEOUT, poll_harness_status
 
 REGION = os.getenv("AWS_DEFAULT_REGION")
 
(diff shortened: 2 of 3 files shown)


evandrofranco (approved):
OK

PR #1864, opened 2026-07-30
fix(01-harness): correct execution limits, share the poller, and stop leaking the role in 4 advanced samples

rmncardoso:
**Issue number**: _to follow_

```
Changes to four advanced Harness samples (01-custom-containers, 03-execution-limits,
04-mcp-integration, 13-aws-skills), applying the shared poller and role-lifecycle
fixes from #1857 and correcting a factual error in the execution-limits demo.
```

Follow-up to #1857 (merged), which added `utils/harness.py`. This applies the same
fixes to the four advanced samples that need no extra infrastructure to run.

### What changed

- **03-execution-limits** — the demo claimed `maxTokens` truncates the response
  (`maxTokens=10` → "~10 tokens"). On `InvokeHarness`, `maxTokens` is a
  **per-iteration** ceiling, not a budget for the whole invocation, so the agent
  keeps going across iterations and the answer is not truncated. Verified live:
  `maxTokens=256` still produced **1413** output tokens and stopped with
  `end_turn`. Corrected the code, docstring and README, and switched the demo to
  compare per-run output tokens. `maxIterations` and `timeoutSeconds` do stop the
  agent (`max_iterations_exceeded` / `timeout_exceeded`), which the demo still
  shows.
- **All four** — replace the local wait-for-READY loop with `poll_harness_status`
  from #1857. The local loops had a fixed ~60s ceiling on an operation that takes
  ~150s (longer for a container pull or an AWS-skills harness), and several
  checked for a `"FAILED"` status that isn't in the enum (`CREATE_FAILED` /
  `UPDATE_FAILED` / `DELETE_FAILED`), so they could never detect a failure.
- **03-execution-limits, 04-mcp-integration** — role and harness creation were
  above the `try`, so a `CreateHarness` failure exited leaving the shared IAM role
  behind. Moved creation inside `try/finally` and made cleanup delete each
  resource independently, matching #1857's getting-started fix.
- **01-custom-containers, 13-aws-skills** — already create inside `try`; gain a
  `created_role` guard so a caller-supplied `--role-arn` is never deleted.

### User experience

Before, on a fresh account these samples either timed out at 60s while the harness
was still `CREATING` and then failed at the next call, or (custom-containers /
visual-testing) failed the `update_harness` step with a `ConflictException`
because the poll fell through early. execution-limits additionally taught the
wrong mental model for `maxTokens`. After, each sample waits for the real status,
raises immediately with the service's `failureReason` on a genuine failure, and
never leaves the shared role behind.

### Testing

Live in `us-west-2` on an account with no pre-existing Harness resources — each
sample provisioned, ran end to end, and cleaned up with no leaked role or harness:

- **01-custom-containers** — attached `node:slim`, agent wrote and ran a Node
  server, installed an npm package on the custom image; both the create and the
  `update_harness` poll reached READY.
- **03-execution-limits** — `maxIterations=1` → `max_iterations_exceeded`;
  `timeoutSeconds=5` → `timeout_exceeded`; `maxTokens=256` → **not** truncated
  (1413 output tokens, `end_turn`), confirming the corrected behaviour.
- **04-mcp-integration** — MCP tools loaded and invoked across Parts 1–2, the
  invalid-URL error path (Part 4) surfaced cleanly, Part 5 ran.
- **13-aws-skills** — `--mode glob` loaded `core-skills/*` and the agent
  enumerated them (this is the sample that needs the boto3 floor from #1857).

`ruff check` and `ruff format --check` clean on all four files.

### DIFF
--- a/01-features/01-harness/01-advanced-examples/01-custom-containers/custom_container.py
+++ b/01-features/01-harness/01-advanced-examples/01-custom-containers/custom_container.py
@@ -36,8 +36,9 @@ import boto3
 
 sys.path.insert(0, str(Path(__file__).parent.parent.parent))
 
+from utils.client import get_agentcore_client, get_agentcore_control_client
+from utils.harness import poll_harness_status
 from utils.iam import create_harness_role, delete_harness_role
-from utils.client import get_agentcore_control_client, get_agentcore_client
 
 # ── Language Presets ───────────────────────────────────────────────────────────
 LANGUAGE_PRESETS = {
@@ -155,6 +156,7 @@ def run_command(harness_arn, session_id, command):
 
 
 harness_id = None
+created_role = False
 try:
     # ── Step 0: IAM role ──────────────────────────────────────────────────────
     print("\n=== Step 0: IAM Role ===")
@@ -163,6 +165,7 @@ try:
         print(f"Using provided role: {role_arn}")
     else:
         role_arn = create_harness_role()
+        created_role = True
         print("Waiting for IAM propagation...")
         time.sleep(10)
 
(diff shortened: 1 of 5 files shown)


evandrofranco (approved):
ok

PR #1866, opened 2026-07-30
fix(01-harness): fix memory lifecycle, share the poller, and stop leaking the role in 3 use-case samples

rmncardoso:
**Issue number**: _to follow_

```
Changes to the three use-case Harness samples (01-travel-agent, 02-webapp-visual-testing,
03-aws-builder-agent), applying the shared poller and role-lifecycle fixes from #1857/#1864
and correcting a cluster of AgentCore Memory lifecycle bugs in the travel agent.
```

### What changed

- **03-aws-builder-agent** — **never deleted the shared execution role at all**; it imported only `create_harness_role`. Every sample in this folder shares the role name `HarnessExecutionRole`, so the leftover role breaks the next sample's run. Now deletes it in `finally`, guarded by a `created_role` flag so a caller-supplied `--role-arn` is never destroyed. Also drops its private poller for the shared one — the local copy checked for a `"FAILED"` status that isn't in the enum (`CREATE_FAILED` / `UPDATE_FAILED` / `DELETE_FAILED`), so it could never detect a real failure, and its 180s ceiling left little margin over the ~145s a harness actually takes.

- **02-webapp-visual-testing** — both wait-for-READY loops were `for i in range(24)` with a 5s sleep: a hard 120s ceiling, while the live run needed ~145s. The loop fell through instead of raising, so the `update_harness` call that attaches the Node container then failed with `ConflictException`. Both replaced with `poll_harness_status`.

- **01-travel-agent — Memory lifecycle.** Four separate bugs in Part 4:
  - `create_memory` / `get_memory` return the resource under a top-level `"memory"` key only, so `resp.get("id") or resp.get("memory", {}).get("id")` had a **dead first branch**. Verified live: `resp.get("id")` is `None`.
  - The Memory status enum is `CREATING → ACTIVE` with a plain `FAILED` and **no `READY`** — different from the harness enum. The old loop waited for `("ACTIVE", "READY")` and never checked `FAILED`, so a failed memory was polled to the ceiling and its `failureReason` never surfaced. New `poll_memory_status()` honours the real enum and raises with the reason.
  - It fell back to **adopting any memory whose id contained `"TravelGuide"`** and then deleted it during cleanup — destroying a resource the sample doesn't own. `memory_id` is now only set when this run created the memory; a pre-existing one raises `ConflictException`, which is caught and the memory demo skipped with a clear message.
  - The harness create-wait (`range(12)`, a 60s ceiling versus ~145s observed) and the post-`update_harness` wait both replaced with the shared poller.
  - Cleanup deletes the harness before the memory it references, each guard independent so one failure can't skip the role deletion.

- Cleared pre-existing lint debt on the touched files (import order, `dict()` → literal, blind-except `noqa`) and removed a decorative shebang (EXE001), matching what #1864 did for `aws_skills.py`.

### User experience

Before, on a fresh account: **aws-builder-agent** completed successfully but left `HarnessExecutionRole` behind, so whichever sample ran next inherited a stale role with nothing pointing at the cause. **webapp-visual-testing** timed out at 120s while the harness was still `CREATING`, then failed the container attach with `ConflictException`. **travel-agent** could invoke a harness that was still `CREATING`, and its Memory step never reported a genuine failure — and if a `TravelGuideMemory` already existed, the sample silently adopted it and deleted it on the way out.

After, each sample waits for the real status, raises immediately with the service's `failureReason` on a genuine failure, only ever deletes resources it created itself, and leaves nothing behind.

### DIFF
--- a/01-features/01-harness/02-use-cases/01-travel-agent/README.md
+++ b/01-features/01-harness/02-use-cases/01-travel-agent/README.md
@@ -83,7 +83,7 @@ Navigate to **CloudWatch → X-Ray → Traces** to see the full agent loop break
 ## Troubleshooting
 
 ### Issue: memory creation fails with `ConflictException`
-**Solution**: A memory named `TravelGuideMemory` already exists. The script handles this by listing and reusing it. If you want a fresh memory, delete the old one first.
+**Solution**: A memory named `TravelGuideMemory` already exists — usually left behind by an earlier run that did not finish cleaning up. The script only deletes memories it created itself, so it will not silently reuse and then delete one it does not own; it prints a warning and skips the memory demo. Delete the leftover memory (`aws bedrock-agentcore-control delete-memory --memory-id <id>`), or re-run with `--skip-memory`.
 
 ### Issue: Part 4 takes too long
 **Solution**: memory provisioning takes 3-5 minutes. Use `--skip-memory` to skip the memory demo entirely.
--- a/01-features/01-harness/02-use-cases/01-travel-agent/travel_agent.py
+++ b/01-features/01-harness/02-use-cases/01-travel-agent/travel_agent.py
@@ -35,8 +35,9 @@ import boto3
 
 sys.path.insert(0, str(Path(__file__).parent.parent.parent))
 
+from utils.client import get_agentcore_client, get_agentcore_control_client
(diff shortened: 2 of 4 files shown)


evandrofranco (approved):
OK

PR #1868, opened 2026-07-31
fix(01-harness): stop 01-custom-containers prompting for tools its images lack

rmncardoso:
## Problem

`01-features/01-harness/01-advanced-examples/01-custom-containers` asks the agent to run three tools that are **not present in the container images the sample itself attaches**, and the README documents the resulting output as the expected behaviour.

| Preset | Image | Sample asks for | Actually in the image |
|---|---|---|---|
| `python` | `python:3.12-slim` | `curl localhost:3000` | no `curl`, no `wget` |
| `go` | `golang:1.24` | architecture "using `'file'` command" | no `file` |

The `go` case is visible in a live transcript. The agent tries `file`, fails, and recovers on its own:

> Now let's check the file size and architecture:
> **Let me try with readelf as an alternative to show the architecture:**

So the demo passes while doing something other than what [README.md](../blob/main/01-features/01-harness/01-advanced-examples/01-custom-containers/README.md) says it does, at the cost of a wasted turn.

The README's Architecture block also advertised "use curl" for *every* image and described the microVM as Node-only, and it documented **no python prompt at all** even though `--language python` is a supported preset.

## Changes

- **`python` prompt** → asks for `urllib` instead of `curl`, naming why.
- **`go` Part 5 prompt** → asks for `ls -lh` + `readelf -h` directly. Shell calls in that step drop from **3 to 2** and the fallback turn disappears.
- **README** → generalised the Architecture block, added a note that minimal images omit common tools, added the missing python prompt, corrected the cross-compile expected behaviour.
- **Bounded retry on the first `invoke_harness`** (3 attempts, 15 s apart, on `RuntimeClientError`). A harness reporting `READY` does not guarantee the container's microVM is answering health checks yet: one python run died with `Runtime health check failed or timed out` while that runtime's own CloudWatch log group recorded a clean startup and **zero** errors. Only the *call* is retried, never a stream that already emitted output, so no text can print twice. A genuinely broken image still fails, on every attempt, and raises.
- **Harness naming follows `--language`** — Go/Python runs no longer create resources labelled `NodeContainer_*`, which made leaked resources and their auto-named managed memory indistinguishable in the console.

## Testing

All three presets deployed, invoked and torn down live in `us-west-2`, each exit 0:

| Preset | Result | Evidence |
|---|---|---|
| `go` | PASS | `readelf` on the first try, no fallback turn; cross-compiled 8,454,144-byte ELF64 x86-64 |
| `python` | PASS | agent used `urllib.request`, HTTP 200, no `curl` attempt |
| `node` | PASS | unchanged default path; node v26.5.1 / npm 11.17.0; chalk installed and ran |

Retry logic additionally unit-tested on four paths: healthy first call (1 attempt, no retry), transient failure then success (output printed exactly once), recovery on the final attempt, and a permanently broken runtime (raises after exactly 3 attempts, no infinite loop).

Teardown verified after every run — harness, inline policy and role all deleted; `HarnessExecutionRole` confirmed absent afterwards.

### DIFF
--- a/01-features/01-harness/01-advanced-examples/01-custom-containers/README.md
+++ b/01-features/01-harness/01-advanced-examples/01-custom-containers/README.md
@@ -36,10 +36,15 @@ By default, harness runs on Amazon Linux 2023 with Python. Custom containers let
                 │
                 ▼
         [Firecracker microVM]
-        - node, npm, full Node.js ecosystem
-        - Agent can install packages, run servers, use curl
+        - Whatever the image ships (e.g. node + npm, or the Go toolchain)
+        - Agent can install packages and run servers
         - Same session ID = same VM state persists
 ```
+
+> **Only what the image ships is available.** Minimal base images omit tools you
+> may take for granted: `python:3.12-slim` has neither `curl` nor `wget`, and
+> `golang:1.24` has no `file`. Prefer language built-ins (Node's `http`, Python's
+> `urllib`) over shelling out, or install the tool first.
 
 ## Sample Prompts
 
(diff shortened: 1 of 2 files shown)


evandrofranco (approved):
ok

PR #1870, opened 2026-08-01
fix(01-harness): correct the gateway docs, harden the pollers, stop leaking the role

rmncardoso:

The `02-gateway-integration` sample works end to end, but the README describes a different sample than the one that ships, the gateway pollers can't report the failures the service actually returns, and every run leaves the IAM execution role behind.

This builds on #1863 (this branch's direct parent), which fixed the *harness* poller in this file. The gateway pollers a few lines above it have the same two defects — a timeout shorter than the documented wait, and status checks against values the API never returns — and neither was in scope there.

**Documentation**

- **A "Create routing rule" step the sample never performs.** A gateway routes to its targets as soon as they report `READY`. Rules are real (`CreateGatewayRule` exists) but are a separate, optional feature for shaping traffic across multiple targets. Replaced with a note saying so.
- **"IAM auth (`NONE` type)" contradicts itself.** `NONE` means *no* inbound authorizer; `AWS_IAM` is a distinct value this sample doesn't use. It also collapsed two independent controls into one — the distinction `07-oauth` gets right, so the two samples disagreed. Replaced with a "Two independent authorization sides" section and an inbound/outbound table.
- **A harness "wired to the gateway's ARN" — no such binding exists.** `CreateHarness` takes no gateway parameter; the ARN travels per-invoke in `tools`. Stated the decoupling explicitly.
- **Step numbers didn't match the output** — README and docstring numbered 1–7, the script prints `Step 0`–`Step 4` plus cleanup.
- **The flow list promised cleanup would delete the IAM role**, which appears nowhere in the code, and the Clean Up snippet omits it too.
- **Three new troubleshooting entries:** HTTP 429 from Exa's shared free tier (with the outbound `API_KEY` config), the `UPDATE_UNSUCCESSFUL`/`SYNCHRONIZE_UNSUCCESSFUL` statuses, and a target stuck in `CREATE_PENDING_AUTH`.

**Code**

- **`GATEWAY_POLL_TIMEOUT` was 120s while the README promised "2-3 minutes."** Raised to 300s, matching `07-oauth`'s budget for these same operations. Both resources actually reach `READY` in ~5s, so the old ceiling only bit on the slow runs it was least able to survive.
- **The pollers checked for statuses the API never returns.** A gateway reports `UPDATE_UNSUCCESSFUL`, not `FAILED`, when an update fails; a target adds `SYNCHRONIZE_UNSUCCESSFUL`. Neither was terminal, so a failed resource was polled until timeout and `statusReasons` never surfaced. The target poller also tested `DELETE_FAILED`, which isn't in the target enum at all — a dead branch. The three `*_PENDING_AUTH` states now fail fast instead of spinning on a state that can't clear by itself.
- **The stream handler named only `internalServerException`.** `validationException` and `runtimeClientError` are modelled members of the same event-stream union, so the handler now names them too. Being precise about what this does and does not do, because my first draft of this description overstated it: with standard AWS exception framing (`:message-type: exception`) botocore sets a 400 and raises `EventStreamError` before the loop ever sees a dict, so these two branches are defensive completeness rather than a fix for a failure I reproduced. The live path for a rejected invoke is the `EventStreamError` catch already in `stream_response`. I was unable to trigger either event against a real harness.
- **`REGION` read `AWS_DEFAULT_REGION` alone**, while `utils/client.py` honours `AWS_REGION` too and treats `""` as unset. A shell exporting `AWS_DEFAULT_REGION=""` built a working `harness_control` and a `gw_control` that died on `Invalid endpoint: https://bedrock-agentcore-control..amazonaws.com`. Now imports the shared `REGION`.
- **`_cleanup` deletes the execution role**, guarded by `created_role` so a role passed via `--role-arn` is never touched — the name is shared by all 8 samples in this folder. Mirrors the guard merged in #1864.

Also annotated the `create_gateway` / `create_gateway_target` call sites with which authorization side each parameter controls, dropped the unused `os` import, and restored alphabetical import order.

### User experience

**Before** — a successful run on unmodified `main`:

```
Cleaning up...
  Deleted harness: GatewayHarness_81cc1fb4-Ow49bI1ek1
  Deleted target: FMBJ1RTMRS
  Deleted gateway: gatewaydemo-f5c5ed82-ilqcro0d78

$ aws iam get-role --role-name HarnessExecutionRole
HarnessExecutionRole    2026-08-01T00:01:09+00:00     <- left behind
```

**After:** the same three deletes, followed by `Deleted inline policy: HarnessExecutionPolicy` and `Deleted role: HarnessExecutionRole`.

A reader following the README no longer looks for a routing-rule step that doesn't exist, and a failed gateway or target stops the run with the service's own reason instead of timing out silently.

### Testing

**Live end-to-end runs in `us-west-2`** — one on unmodified `upstream/main` to capture the shipped behaviour above, one on the exact bytes committed here (md5-verified before and after). Every run: gateway and target `READY`, the agent called `exa-search___web_search_exa` through the gateway and returned real results, exit 0. Post-run sweeps confirmed a clean account, including the auto-provisioned managed memory, which drains on its own.

**Unit-checked against stubbed clients**, since most new branches can't be reached from a healthy account: both pollers (happy path, every terminal and pending-auth status, transient passthrough, timeout); the stream handler (tool-use, raw mode, `tools` payload shape, and each error event injected as a plain dict); `_cleanup` (`created_role` true/false/default, and failure before any resource exists). The empty-region regression was reproduced before the fix and confirmed resolved after.

Worth a reviewer's attention: the two new stream-error branches are exercised only by injected dicts. Against the real service these exceptions arrive as `:message-type: exception`, which botocore turns into a raised `EventStreamError`, so the branches are unreachable on that path. They are cheap and consistent with the other samples in the folder, but I would rather flag that than let the diff imply I had reproduced them.

**Every API-shape claim above was asserted against the botocore 1.43.59 service model** — both enums for authorization, `apiKeyCredentialProvider` members, both status enums (including `DELETE_FAILED`'s absence from the target enum), the `InvokeHarness` stream events, and the absence of any gateway parameter on `CreateHarness`.

Both CI gates pass. Ruff findings on this file go from 3 to 0 — the `I001` came from #1863's own new import, and the two `BLE001`s were already failing on `main`.

### DIFF
--- a/01-features/01-harness/01-advanced-examples/02-gateway-integration/README.md
+++ b/01-features/01-harness/01-advanced-examples/02-gateway-integration/README.md
@@ -18,30 +18,40 @@ discovers and calls tools via the gateway.
 ## What is AgentCore gateway?
 
 AgentCore gateway is a managed proxy between your agent and external tool servers (MCP, HTTP).
-It provides centralized auth, routing rules, and observability for all tool traffic — without
+It gives you one place to handle auth, routing and observability for all tool traffic — without
 changing your agent code.
 
 ```
 [harness] → tools=[{type: "agentcore_gateway", gatewayArn: ...}]
                 │
-                ▼
-            [gateway] ── auth + routing + observability
+                ▼   inbound auth: authorizerType (NONE here)
+            [gateway] ── routing + observability
                 │
-                ▼
+                ▼   outbound auth: credentialProviderConfigurations (unset here)
             [MCP Target] ── external MCP server
 ```
 
 ## End-to-End Flow
 
+The step numbers match the ones the script prints:
+
 ```
-1. Create IAM execution role   (reuses utils/iam.py helper)
-2. Create gateway              → IAM auth + MCP protocol
-3. Add MCP target              → remote MCP server endpoint (default: Exa)
-4. Create routing rule         → directs traffic to the target
-5. Create harness              → wired to the gateway's ARN
-6. invoke_harness              → agent discovers tools via gateway and calls them
-7. Cleanup                     → delete harness, target, gateway, IAM role
+Step 0. Create IAM execution role   (reuses utils/iam.py helper)
+Step 1. Create gateway              → MCP protocol, no inbound authorizer
+Step 2. Add MCP target              → remote MCP server endpoint (default: Exa)
+Step 3. Create harness              → plain harness; it holds no gateway reference
+Step 4. invoke_harness              → passes the gateway ARN in `tools`, agent calls them
+Then    Cleanup                     → delete harness, target, gateway, IAM role
 ```
+
+The harness and the gateway are not bound to each other. `create_harness` takes no
+gateway parameter — the gateway ARN travels per-invoke in the `tools` argument to
+`invoke_harness`, so one harness can be pointed at different gateways on different
+calls, and deleting either resource has no effect on the other.
+
+A gateway routes to its targets as soon as they are `READY`, so no routing rule is
+needed to run this sample. Rules (`CreateGatewayRule`) are a separate, optional
+feature for shaping traffic across multiple targets.
 
 ## Sample Prompts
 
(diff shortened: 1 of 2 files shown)


evandrofranco (approved):
ok

PR #1876, opened 2026-08-01
fix(01-harness): stop 05-agent-skills reporting success it never checked, and leaking on failure

rmncardoso:

The `05-agent-skills` sample cannot finish a run on a fresh account, and when it fails it leaves billable resources behind. Underneath that, it prints four green ticks after commands whose exit codes are never read, and states a sheet count for a file nothing ever opened.

**Code**

- **Two hand-rolled polls capped waiting at 60s and fell through silently.** Both were `for i in range(12)` with a 5s sleep and — verified on the shipped bytes — **no `else` clause and no `raise`**, so a timeout wasn't an error but a silent fall-through into the next call. Measured create time in a real account is **~95-133s**, and this sample's own `utils/harness.py` documents ~150s public / ~255s VPC, so the ceiling sat below the operation it waited on. The very next statement, `update_harness`, then ran against a `CREATING` harness. Now uses the shared `poll_harness_status` — 600s, raising on timeout and on `CREATE_FAILED`/`UPDATE_FAILED`/`DELETE_FAILED` with the service's `failureReason`.
- **The file had zero `try` blocks, so that crash leaked.** Cleanup was straight-line code at the bottom, past four unprotected lifecycle calls, so anything raising before it skipped it — orphaning a billable harness *and* `HarnessExecutionRole` with its inline policy. That role name is shared by every sample in this folder, so the leftovers also make the next run of any neighbour behave unpredictably. Now one `try/finally` deleting each resource independently: `harness_id` is `None` if `CreateHarness` itself failed, and a `delete_harness` error must not skip the role deletion.
- **`✅ xlsx skill installed` printed unconditionally**, directly after a command whose exit code was never read. On a non-Debian image `apt-get` returns **127**; with a bad package name, **100**. The tick printed either way. Compounding it, stdout and stderr were concatenated and only the last 500 chars shown, so an early `apt-get` failure scrolled out of view. The exit code is now checked, the two streams print separately, and a failed install raises — every later part depends on it, so continuing is never useful.
- **The verification result was computed, printed, and thrown away.** The sample ran `ls .agents/skills/xlsx/ && echo 'skill directory OK' || echo 'skill not found'` — exactly the right question — with **no `if`** anywhere after it, so a VM printing `skill not found` went straight on to `invoke_harness(..., skills=[{"path": ".agents/skills/xlsx"}])`, turning a clear setup error into a confusing model response two parts later. Now gated on the string the command was written to emit.
- **All three retrievals discarded stderr and the exit code.** Each ran `... 2>/dev/null` and accumulated only `delta["stdout"]`; `contentStop` and `exitCode` appear **zero times** in the shipped file, though `InvokeAgentRuntimeCommand` streams all three signals. A missing file and an unreadable one produced the identical empty string, so the sample couldn't say *why*. Added a `run_command` helper returning `(stdout, stderr, exit_code)`.
- **A workbook was reported for bytes that were not one.** The gate was `if b64_data.strip():` — any non-empty stdout — and the sheet count was a string literal, `"   Sheets: 3 (Summary, Monthly Breakdown, Top Products)"`. Nothing opened the file. An `.xlsx` is a zip container, so `zipfile.is_zipfile` now rejects non-workbooks dependency-free and the count is read from the `xl/worksheets/sheet*` entries. The file is kept on disk when it fails the check, so the reader can inspect what actually arrived.

**Documentation**

- **The three command outputs are now documented** under Key Concepts with the stream-loop snippet, since reading only stdout is what made every false success above possible.
- **`node:slim` and LibreOffice.** The skill's recalc step needs `soffice`, which the Node image doesn't ship — but the agent notices and `apt-get install`s LibreOffice mid-run, so delivered workbooks *are* fully recalculated (`docProps/app.xml` names LibreOffice as the generating application). The real cost is latency on a fresh session, and attaching an image that already includes `soffice` avoids the download.
- **New troubleshooting entries** for the `ConflictException`, the corrupted-download message and the skill-install failure, so none of the new diagnostics reads as a bug.

### User experience

**Before** — unmodified `main`, fresh account:

```
  [10] CREATING
  [11] CREATING
  [12] CREATING
Traceback (most recent call last):
  ...
botocore.errorfactory.ConflictException: An error occurred (ConflictException) when
calling the UpdateHarness operation: Cannot update agent SkillsDemo_698c89f2 while
it is CREATING. Wait and try again.
```

The run ends there — no `=== Cleanup ===`, so `SkillsDemo_698c89f2-aKeMNZXGmj` and `arn:aws:iam::…:role/HarnessExecutionRole` are both left behind. Note the trap: you cannot `delete_harness` while it is `CREATING`, so cleaning up by hand means polling to `READY` first.

**After** — same account, same conditions:

```
  [19] READY
✅ Harness is ready
...
skill directory OK
✅ xlsx skill installed
...
   Sheets: 1
   Sheets: 3
Deleted harness: SkillsDemo_06208840-nFv9S0rKbD
Done.
```

`READY` arrived at poll 19, past where the old ceiling gave up.

**The worst false success**, driven on the same corrupt file at the same path — before, then after:

```
✅ Saved to /tmp/probe_d_corrupt.xlsx (36 bytes)
   Sheets: 3 (Summary, Monthly Breakdown, Top Products)
   -> is_zipfile = False        # both lines above were false
```

```
⚠️  Saved /tmp/probe_d_corrupt.xlsx (36 bytes) but it is not a valid .xlsx.
   First bytes: b'this is definitely not a spreadsheet'
```

A missing file now says why instead of going quiet:

```
⚠️  Budget spreadsheet not retrieved (exit 1).
   base64: /tmp/amsterdam_budget.xlsx: No such file or directory
   The agent may have answered inline instead of writing the file.
```

### Testing

**Live in `us-west-2`, on unmodified `main` first** to capture the shipped behaviour above. From a verified-clean baseline — no `HarnessExecutionRole`, no `SkillsDemo_*`, no managed memories — the leak reproduced end to end, and the orphans were cleaned up by hand afterwards.

**Then live end to end on the exact committed bytes** (md5-verified before the run, after it, and against `git show HEAD:`), `EXIT=0`: all parts ran, the container update polled `UPDATING`→`READY`, install exited 0 with the real skill directory listed, `Sheets: 1` and `Sheets: 3` were both read from the files, and cleanup deleted harness, inline policy and role. The post-run sweep was clean apart from the service's auto-provisioned managed memory, which drains on its own.

**Both delivered files were validated independently of the sample:** `is_zipfile` true and `testzip()` CRC-clean on each; 1 and 3 worksheet parts, matching what the script reported; sheet names `['Amsterdam Budget']` and `['Summary', 'Monthly Breakdown', 'Top Products']`; and 52 and 15 formulas respectively, each carrying its cached result (`=D2*1.1` → `132`, `=IF(D4>=0,…)` → `'Below Target'`). Part 4's sheet names match what the old code hardcoded, but they are now a measurement rather than a guess.

**Old and new logic were driven back to back on one VM in one session**, so no difference is attributable to a different environment: the discarded `skill not found` (model invoked anyway, now raises); `exit 100` on a bad package (tick, now `RuntimeError`); a missing file (`⚠️ No file generated` with no reason, now `exit 1` plus `No such file or directory`); and the corrupt-download pair shown above. A genuine 3-sheet workbook written to that same path still passes (`Sheets: 3`, `testzip()` `None`), which rules out the check being merely stricter, and a truncated zip — valid `PK\x03\x04` header, cut mid-stream — is also rejected.

**82 unit checks**, all passing, for branches a healthy account can't reach. The functions under test are extracted from the shipped file by AST and exec'd, so the checks run the real bytes rather than a retyped copy — the module cannot simply be imported, since it creates a harness at import time. Covered: `run_command` accumulating all three signals across chunks, a missing `stdout` key, an absent `contentStop`, non-chunk events; `download_xlsx` across missing file (replaying the exact stderr and exit code captured live), exit 127, non-zero exit with empty stderr, not-base64, bad padding, not-a-zip, an error-text payload, a truncated zip, line-wrapped base64, and valid workbooks at 1, 3 and 5 sheets; both Part 2 gates and that the tick prints after them; the `finally` block with `harness_id=None` and with `delete_harness` raising; and the structural invariants — no lifecycle call outside the `try`, two `poll_harness_status` calls, no `range(N)` poll left, the hardcoded sheet claim gone.

Both CI lint gates pass on the changed file.

### DIFF
--- a/01-features/01-harness/01-advanced-examples/05-agent-skills/README.md
+++ b/01-features/01-harness/01-advanced-examples/05-agent-skills/README.md
@@ -65,6 +65,34 @@ client.invoke_harness(
 
 **File download**: Generated files live on the agent's remote VM. Use `invoke_agent_runtime_command` with `base64` to transfer them locally.
 
+**Read all three outputs of a command**: `invoke_agent_runtime_command` streams `stdout`,
+`stderr` *and* `contentStop.exitCode`. Check the exit code before treating the output as
+good — a command that failed still streams whatever it managed to write, so stdout alone
+cannot tell success from failure:
+
+```python
+for event in resp["stream"]:
+    if "chunk" in event:
+        chunk = event["chunk"]
+        if "contentDelta" in chunk:
+            stdout += chunk["contentDelta"].get("stdout", "")
+            stderr += chunk["contentDelta"].get("stderr", "")
+        elif "contentStop" in chunk:
+            exit_code = chunk["contentStop"].get("exitCode")
+```
+
+**`node:slim` has no LibreOffice, and the skill wants one**: the xlsx skill recalculates a
+finished workbook so that each formula also carries its *cached result* — the value
+`openpyxl(data_only=True)` and pandas read without opening Excel. That step needs
+LibreOffice (`soffice`), which the Node image does not ship. In practice the agent notices
+and `apt-get install`s LibreOffice mid-run, so the downloaded files do come back fully
+recalculated (`docProps/app.xml` names LibreOffice as the generating application). It works,
+but it is a large package to install on every fresh session — expect the first spreadsheet
+prompt to take noticeably longer. If the install is skipped or fails, the formulas are still
+correct; only their cached values are missing, and `data_only=True` reads `None` until Excel
+or LibreOffice opens the file. Attach an image that already includes `soffice` to avoid the
+download entirely.
+
 ## Troubleshooting
 
 ### Issue: `npx skills add` command hangs
@@ -73,8 +101,24 @@ client.invoke_harness(
 ### Issue: Skill not found error during invocation
 **Solution**: Verify the path exists: `ls -la .agents/skills/xlsx/`. Run the installation in the same session you're invoking from.
 
+### Issue: `ConflictException: Cannot update agent ... while it is CREATING`
+**Solution**: The harness was not READY before the next call was made. Creating a harness
+takes ~150s on the public network and longer when a container image has to be pulled on top,
+so poll to `READY` before calling `update_harness` or running a command. Use
+`poll_harness_status` from `utils/harness.py`, which raises on timeout and on
+`CREATE_FAILED`/`UPDATE_FAILED` instead of falling through silently.
+
 ### Issue: Generated xlsx file is empty or corrupted
-**Solution**: Check that the agent's file path matches the base64 read command. The file must exist on the VM before you can download it.
+**Solution**: Check that the agent's file path matches the base64 read command. The file must
+exist on the VM before you can download it. This sample validates the transfer: it reports the
+`base64` exit code and stderr when the file is missing, and verifies that what arrived is a zip
+container (an `.xlsx` is a zip) before reporting a sheet count read from the file itself.
+
+### Issue: The script reports the skill install failed
+**Solution**: Read the command output above the error. The install runs
+`apt-get update && apt-get install git -y && npx skills add ...`, so it needs a Debian-based
+container (the `node:slim` image this sample attaches in Part 1) and outbound network access.
+The script stops here deliberately — every later part depends on the skill being present.
 
 ## AgentCore CLI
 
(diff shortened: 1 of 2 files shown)


evandrofranco (approved):
ok

PR #1918, opened 2026-08-05
Add the AWS Agent Registry preview-to-GA migration tool

sg-nitd:
Adds a migration tool under `01-features/07-centralize-and-govern-your-ai-infrastructure/03-registry/04-migrate-to-new-namespace`.

AWS Agent Registry moves out of the `bedrock-agentcore` namespace into its own `agent-registry` namespace at GA, and the registry and registry-record schemas change with it. This tool migrates the data: it extracts existing registries and records, transforms them to the GA schema, and loads them into the new namespace in the same account and region.

Extract and load are separate, idempotent steps. Nothing is written to GA without `--live`, so a run can be inspected before anything is created, and re-running after a failure retries only what did not succeed.

Two ways to run it:

- **One-time migration** — local or CloudShell, no infrastructure to deploy.
- **Managed migration with AWS Glue** — a CDK stack of two Glue jobs and an S3 staging bucket, for unattended runs and the incremental load at cutover.

Both control-plane service models come from `boto3`. Reports cover per-mapping counts, each failed record with its error, and the old-to-new record ID crosswalk.


mchaitra007 (approved):
Reviewed the code on gitlabl, provided feedback there, all are incorporated. Migration code looks good.

PR #1931, opened 2026-08-07
Run the Agent Registry migration Glue jobs on Glue 5.0 glueetl, and drop GA terminology

sg-nitd:
Follow-up to #1918, with two changes: the deployed AWS Glue jobs move to a runtime that can install the SDK they need, and every "GA" reference is removed from the sample.

## The jobs had to change runtime

Both control-plane clients are modeled `boto3` operations, and neither service model lives in this repository — they come from the worker's installed SDK. No AWS Glue image carries `agent-registry-control`: the Python shell runtime is boto3 1.21 and the Glue 5.0 runtime is boto3 1.34.

Installing it on a Python shell job is not possible, for reasons that are all upstream facts rather than preferences:

| Fact | Consequence |
| --- | --- |
| `agent-registry-control` first shipped in botocore **1.43.66** | nothing older has the model |
| every botocore from **1.43.0** onward requires Python >= 3.10 | the model implies 3.10+ |
| the last botocore allowing Python 3.9 was **1.42.97**, which predates the model | no pin satisfies both |
| a Glue **Python shell** job is Python 3.9.10 at GlueVersion 3.0, 4.0 **and** 5.0 | raising GlueVersion does not help |
| `CreateJob` rejects any `command.pythonVersion` outside `^([2-3]\|3[.]9)$` | the version cannot be overridden |

A `glueetl` command on Glue 5.0 is Python 3.11, which is the only Glue runtime `boto3==1.43.66` installs on. Both jobs remain single-threaded boto3 scripts and never create a `SparkContext`.

## Terminology

The service is not generally available, so the sample no longer describes it that way. Documentation, source, tests and CLI output are all covered:

- Identifiers and settings: `GaRegistryClient` -> `TargetRegistryClient`, `validate_ga_request` -> `validate_target_request`, `iam.gaWriteActions` -> `iam.targetWriteActions`, `approval.gaStatusCounts` -> `approval.targetStatusCounts`, `api.ga` -> `api.target` in the adapter contract.
- Placeholders and paths: `<ga-registry-id>` -> `<new-registry-id>`, `ga-registry-payloads/` -> `new-registry-payloads/`.
- Prose: the destination is the **target registry** — the vocabulary the code already used for the source/target pair — and the service itself is **the new version of AWS Agent Registry**.
- The README opening no longer announces a general-availability launch date.

Renaming the adapter's section key changes the replay fingerprint. All three computations (checkout, package, deployed) agree on the new value.

## Other changes

- Both jobs become `glueetl` on Glue 5.0, with the SDK pinned exactly in `GLUE_SDK_MODULES` rather than floored — so two runs of one cutover cannot stage and load with different SDKs.
- Sized at the AWS minimum for a batch job, `G.1X` x 2 (`G.025X` is streaming-only). `engine.glueMaxCapacity` is removed, since `MaxCapacity` and `WorkerType`/`NumberOfWorkers` are mutually exclusive; `config.ts` now rejects the removed key by name instead of silently ignoring it, and the fractional-DPU cap on `loadConcurrency` goes with it.
- Glue bookmarks disabled explicitly. An incremental run resumes from the watermark in the staging bucket, which an operator can read and override with `--changed-after`; an opaque bookmark could only disagree with it.
- `AwsSolutions-GL1`/`GL3` suppressed with justifications — Glue logs are run metadata rather than record content, and no bookmark data exists to encrypt.
- A missing service model is now a pre-flight failure (`check_sdk_models`) with an actionable remedy, instead of `UnknownServiceError` from the first client build. The local interpreter is probed **by capability, not by version**, so an operator with a side-loaded model still works.
- `requirements-dev.txt` moves to `boto3>=1.43.66`. The old `<1.40` ceiling existed only to hold CI on Python 3.9 to match the Python shell runtime; that reason is gone, and the pin was breaking virtualenv setup.
- Renames the dual-writes migration pattern to **active-active**, per review feedback on #1918.

## Verification

Deployed and run end to end against real registries in a test account, on both engines:

| | Local | Glue |
| --- | --- | --- |
| `check` | PASS | PASS |
| `extract` | 34 records | same 34, 60s |
| `load --dry-run` | dryRun=34 | dryRun=34, 43s |
| `load --live` | unchanged=34 | unchanged=34, 35m44s |

- Worker log confirms the runtime: `python3.11 -m venv --clear /tmp/glue_venv`, then `pip install boto3==1.43.66 botocore==1.43.66`.
- Deployed job properties read back as `glueetl` / `PythonVersion: 3` / `GlueVersion: 5.0` / `G.1X` x 2.
- Glue and local staged output differ **only** in `runId` and `extractedAt` — no record content differs.
- The Glue live load began with an empty S3 id map, independently rebuilt the index by scanning the target registry, and produced an id map identical to the local one across all 34 source/target pairs.
- After the terminology change: 489 Python tests pass, `tsc --noEmit` clean, `ruff check` and `ruff format --check` clean, `verify:lib` and `verify:fingerprint` pass, and both example configurations synth (the customer-managed-role config still produces no IAM resources).
- Every internal documentation link and anchor re-checked after the renames.

Note: no record was *created* through Glue in this run, because the fixture was already fully migrated — the create branch resolved to "nothing to do". The write path itself is exercised (real client, real target scan, real `--live` decisioning, id map and watermark committed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mchaitra007 (approved):
Switch to GlueETL looks good, all references to GA has been removed.

PR #1947, opened 2026-08-10
Add POC Validator agent sample

schinchli:
## Summary

Adds a new sample under `02-use-cases/02-workflow-automation-agents/poc-validator-agent/`.

Reviews a proposed AWS architecture against segment and industry rules, checks that the services actually connect, prices it, scores the Scope of Work against fixed criteria, and recommends further reading from AWS sources only — then, optionally, runs behind a small web front end so a reviewer can upload a real SOW and hand back a view-limited, expiring share link, without a login for the recipient. Two further optional phases let a reviewer ask a plain-language what-if pricing question, answered by model-authored code executed in a real sandbox rather than a number the model states directly, and search a curated FAQ of recurring findings by vector retrieval.

**AgentCore features used:** Runtime, Memory (SEMANTIC + SUMMARIZATION + USER_PREFERENCE, split explicitly by short-term/long-term scope), Gateway (real Lambda-backed AWS Documentation MCP target), Identity (M2M OAuth via Cognito), Policy Engine (Cedar, `ENFORCE` mode), Evaluations, Observability, Code Interpreter (AWS-managed sandbox, what-if pricing). A Knowledge Base (FMKB) is also used for shared FAQ search; it is documented but not claimed as a rubric point, since Knowledge Base is not one of the rubric's 13 named features.

**Design principle carried through the whole sample:** the model is used only where judgement is needed — reading a diagram, banding SOW prose, authoring a what-if calculation. Validation, pricing, and recommendations are deterministic Python, so nothing a reviewer would take at face value is generated. If a model call fails, SOW scoring and what-if pricing both degrade to a deterministic floor or a clear "unavailable" result and say so, rather than failing the review — observed live against a real deployment, not just unit-tested.

Eleven ADRs in `docs/decisions/` capture the reasoning behind each of these choices.

## Verified before submission

- **Deployed and exercised end to end** in a real AWS account across every phase, including the two optional ones, both by full-agent invocation and by calling the underlying AWS APIs directly and independently of the model-gated steps.
- **Two real defects caught by direct verification, not assumed from documentation**: `InvokeCodeInterpreter`'s response is an event stream, not the flat dict the API reference implies; AgentCore-managed knowledge bases require `managedSearchConfiguration`, not the `vectorSearchConfiguration` shown in standard Bedrock Knowledge Base tutorials. Both fixed and re-verified.
- **Security review**: `bandit`, `pip-audit`, `npm audit`, a repository-wide secrets scan, and a manual IAM review of every grant script and CDK construct. Two findings fixed — an XXE exposure in diagram parsing (now `defusedxml`), a non-constant-time credential comparison (now `hmac.compare_digest`). One dependency advisory remains, documented rather than hidden: a HIGH-severity `brace-expansion` finding bundled inside `aws-cdk-lib` itself, unreachable by `npm overrides` because it ships as a bundled dependency — confirmed against `aws-cdk-lib`'s `bundleDependencies` field and against its latest published release. No wildcard IAM resources anywhere in scope; no hardcoded credentials or account identifiers in any tracked file.
- **106 tests, offline, no AWS account** (`python -m pytest tests/ -q`) — **100% line coverage on `app/pocvalidator/core/*.py`**, the deterministic, offline-testable core (was 72%). `main.py`, `memory/session.py`, and the AWS-calling tool modules are covered by live-deployment verification rather than mocks, consistent with this sample's existing no-mocking convention. A CDK synthesis test suite was added for `web/cdk/lib/web-stack.ts`, which previously had none.
- **`ruff check` against this repo's own `pyproject.toml`**: clean on every new/changed file.
- **Self-assessed against `02-use-cases/use-case-assessment.md`'s own rubric** in `docs/self-assessment.md`: 17 (`UPDATE` tier).
- **Architecture diagram** (`architecture.png`) regenerated with the `diagrams` Python package to include Code Interpreter and the Knowledge Base.

## Test plan

- [x] `python -m pytest tests/ -q` — 106 passed, from a fresh clone of this branch; 100% coverage on `app/pocvalidator/core/*.py`
- [x] `ruff check 02-use-cases/02-workflow-automation-agents/poc-validator-agent/` against the repo's `pyproject.toml`
- [x] `agentcore validate` → `Valid`, against a real deployment in a separate AWS account
- [x] Real `agentcore invoke` / `InvokeAgentRuntime` calls covering every phase, including both optional ones, and the graceful model-fallback path firing for real
- [x] Direct API calls (`StartCodeInterpreterSession`/`InvokeCodeInterpreter`, `bedrock-agent-runtime:Retrieve`) against the deployed resources, independent of the full agent flow
- [x] `bandit`, `pip-audit`, `npm audit`, and a manual IAM/secrets review across the full diff
- [x] `npx jest` in both `agentcore/cdk/` and `web/cdk/`
- [x] `python3 diagrams.py` regenerates `architecture.png` cleanly


schinchli:
Follow-up commit: a word-for-word review before merge caught that the README/self-assessment implied the two custom Evaluators deploy by default — they never have, in this account (`CreateEvaluator` fails on the grading model, most recently reconfirmed on a full teardown-and-redeploy while preparing this PR). Fixed `agentcore.json.template` to ship with `evaluators: []` to match what `./deploy.sh` actually deploys, and updated the docs to state this plainly instead of implying it's live. 67 tests still pass; the architecture diagram was re-verified accurate (never evaluator-specific).

schinchli:
Second follow-up: found something more serious doing a genuine fresh-clone test (not just re-running tests in my already-populated working directory, which was hiding this). The repo's root `.gitignore` was silently dropping several files this sample needs — most critically `web/cdk/lib/web-stack.ts` (the actual CDK stack) and both Dockerfiles (without which `agentcore deploy` can't build the Runtime image). Fixed with negation entries in the root `.gitignore`, following the exact pattern already used there for `visa-b2b-account-payable-agent` and `event-driven-claims-agent`. Re-verified for real this time: fresh `git clone` of this branch into a clean directory → 67/67 tests pass, `web/cdk` installs and compiles cleanly, `architecture.png` is a valid 2162×1407 PNG. Sorry for the churn — better to catch this before review than after.

schinchli:
Follow-up feature: added AgentCore Memory's USER_PREFERENCE strategy (third strategy alongside the existing SEMANTIC + SUMMARIZATION) so a repeat reviewer's inferred preferences persist across sessions. Deployed and verified live (Memory now reports all three strategies, real invocation succeeded post-deploy). The part worth calling out: every Memory namespace here is keyed by actor_id, and the web layer's Lambda had every visitor hardcoded to the same fake identity — which would have made USER_PREFERENCE silently blend everyone's preferences together. Fixed with a per-browser client-side id (localStorage) instead of a login system, documented in ADR 0009 including why that's a deliberate, disclosed simplification rather than real auth.

schinchli:
Pushed two additional AgentCore features, both deployed and verified against a real account before this comment:

**Code Interpreter — what-if pricing (Phase 6a, optional).** A reviewer's plain-language cost question ("what if we dropped Multi-AZ") is turned into a Haiku-authored `compute(lines)` function and executed in AgentCore Code Interpreter's AWS-managed sandbox against the real cost line items — never a number the model just states. See [ADR 0010](https://github.com/schinchli/agentcore-samples/blob/add-poc-validator-agent/02-use-cases/02-workflow-automation-agents/poc-validator-agent/docs/decisions/0010-code-interpreter-for-what-if-pricing.md). The sandbox call itself was verified directly, independent of the code-authoring model step (which degrades gracefully under this account's existing Bedrock Marketplace restriction, same as SOW grading elsewhere in this sample) — and that direct test caught a real bug: `InvokeCodeInterpreter`'s response is an EventStream, not the flat dict the API reference page implies.

**Shared FAQ search — Knowledge Base / FMKB (Phase 6b, optional).** Recurring architecture-review findings (HIPAA/VPC, RDS storage class, Multi-AZ, WAF, and more) are grounded in a curated `AgentCoreKnowledgeBase`, queried with plain `bedrock-agent-runtime:Retrieve` rather than `RetrieveAndGenerate` — so this path has no generation-model dependency at all. See [ADR 0011](https://github.com/schinchli/agentcore-samples/blob/add-poc-validator-agent/02-use-cases/02-workflow-automation-agents/poc-validator-agent/docs/decisions/0011-shared-faq-knowledge-base-not-per-actor-memory.md) for why a shared Knowledge Base fits this better than AgentCore Memory's per-actor namespace model. Verified end-to-end in the live agent, including catching a second real bug: AgentCore-managed knowledge bases reject `vectorSearchConfiguration` and require `managedSearchConfiguration` instead.

Both needed manual IAM grants (and, for the knowledge base, an explicit env var) — confirmed by inspecting the deployed stack that the CDK L3 construct doesn't auto-provision either the way it does for Memory and Gateway. `scripts/grant_code_interpreter_access.sh` and `scripts/grant_faq_knowledge_base_access.sh` close those gaps and are now part of `deploy.sh`.

Test suite is up to 69 (2 new), and the self-assessment's AgentCore feature count moved from 7 to 8 (Code Interpreter is one of the rubric's 13 named features; the Knowledge Base addition is documented but not claimed as a rubric point since it isn't one of the 13).

PR #1982, opened 2026-08-21
feat(use-cases): add GPU music production agent sample

rmncardoso:

```
Adds a use-case sample under 02-use-cases/02-workflow-automation-agents/ in which
three collaborating agents produce and vet a music track inside one AgentCore
Runtime Instances session, because the repo has no sample that runs a model on a
GPU or shows several runtimes sharing one instance and one volume.
```

Three agents, one `runtimeSessionId`, one `g6.xlarge`:

| Agent | Artifact | Compute | Role |
|---|---|---|---|
| `composition_agent.py` | container (ECR) | **GPU** — NVIDIA L4 | A Bedrock model writes a structured brief; **ACE-Step v1 3.5B** (Apache-2.0) then renders real audio locally on the instance GPU |
| `mastering_agent.py` | container (ECR) | CPU | Measures the mix, hands **the measurements** to the model, which returns an EQ/compressor/limiter chain; the DSP applies it and the output is measured again |
| `compliance_agent.py` | **zip on S3** | CPU | Re-measures independently, screens against a generated back-catalogue using chroma features and subsequence DTW across all 12 transpositions, and **computes** the verdict rather than asking a model for it |

What it demonstrates that no existing sample does:

- **A GPU instance type on a capacity provider.** The three existing capacity-provider samples (`03-mcp-`, `08-http-`, `09-a2a-ec2-capacity-provider`) all default to `m6g.large`. This is the first GPU sample in the repo.
- **Local model inference on a persistent volume.** A CUDA torch stack is 3.13 GB of wheels against a 2 GB image cap, so the 14.6 GB stack is built onto a capacity provider EBS volume by a `mode=prepare` step and invoked as a subprocess. That volume survives stop/resume.
- **Collocation as the coordination primitive.** All three runtimes are invoked with the same `runtimeSessionId`, so they land on one instance and pass real audio files through `/mnt/tracks` — no queue, no database, no orchestrator.
- **Mixed artifacts on one capacity provider.** Two container images plus one zip-on-S3 code artifact, which also surfaces that the two artifact types run as different Linux identities on the shared mount.
- **An agent invoking another agent's runtime.** When the screen objects, the compliance agent calls the composition runtime with its own session id, so remediation happens on the instance it is already running on.
- **A computed verdict.** `passed` is arithmetic on the measurements; the model only writes the explanation and is not shown the distances or thresholds at all.

### User experience

**Before.** Nothing in the repo runs a model on a GPU, and nothing shows more than one runtime sharing an instance. A builder asking "can I host my own model on AgentCore, and how do several agents hand large artifacts to each other?" has no worked example — the capacity-provider samples are single-agent and CPU-only.

**After.** Four commands produce audio you can play:

```bash
export AWS_REGION=us-east-2
python scripts/deploy.py                    # creates everything, launches nothing
python scripts/invoke.py --with-catalogue   # ~7 min; the GPU starts billing here
python scripts/cleanup.py                   # deletes the SESSION - that is what stops it
```

The run prints the evidence rather than asserting it:

```
  3. compose (renders audio on the GPU)
    rendered : NVIDIA L4 in 8.98s (peak VRAM 7.63 GiB)
    audio     20.062s  48000Hz  2ch  -7.5 LUFS  peak 0.42 dBTP

  4. master (real DSP, verified by measurement)
    read     : composition.wav  <- written by another agent
    before    -7.5 LUFS  peak 0.42 dBTP
    after     -14.0 LUFS  peak -3.2 dBTP
    targets  : loudness met, true peak held

  5. compliance screen
    verdict  : REVIEW REQUIRED
    screen   : 2 reference(s), closest catalogue_00.wav distance 0.0665 (review)

  -- collocation --
    All 5 steps were served by one instance, as intended.
```

You end with `composition.wav`, `master.wav` and three markdown reports in
`runs/<track>/`. The mastering report records what the model decided *and what it
deliberately left alone* — on the verified run it placed a highpass at 30 Hz to
remove a DC offset it had measured at 0.001171, and bypassed the compressor
entirely because the loudness range was only 0.8 LU.

**Cost is explicit.** `deploy.py` creates no instances; billing starts at the first
invoke and stops when the **session** is deleted, not when a runtime is stopped.
`cleanup.py` deletes the session first for exactly that reason.

### Testing

Deployed and run end to end on a live `g6.xlarge` in `us-east-2`, then torn down and
verified clean (0 instances, 0 orphaned volumes, no roles or buckets left). Confirmed
by measurement, not by the agents' own reports: the audio was re-measured locally and
matched (`-7.5 → -14.0 LUFS`, `+0.42 → -3.2 dBTP`), and the raw render's 9 clipped
runs at full scale became 0 in the master. The compliance zip came in at 82.0 MB
compressed / 235.0 MB uncompressed against the 250 MB / 750 MB limits.


evandrofranco (approved):
ok

PR #1997, opened 2026-08-27
Add OpenAI Agents SDK payment examples to tutorials 01 and 02

mccartnick:
## Summary

Add OpenAI Agents SDK x402 payment examples to Tutorial 01 (local execution) and Tutorial 02 (AgentCore Runtime), using GPT-5.5 through Amazon Bedrock. The Runtime uses a caller-created spending session and documented payment/Bedrock permissions; agent code does not modify IAM.

The payment tool validates and pins public HTTPS destinations while preserving TLS hostname verification, generates one payment proof, and replays the GET once. It reports uncertain payment outcomes without automatically creating another payment. The examples return the agent's final answer, disable the default OpenAI trace exporter, expose a typed GET-only tool schema, and share model defaults.

## Validation

- September 17, 2026: all 50 existing offline regression tests passed on Python 3.12 before removing the test suite, its dedicated CI workflow, and the README test commands.
- Ruff lint, formatting, Python syntax checks, and `git diff --check` pass. The agent, payment tool, and Runtime Python sources are unchanged from the previously validated commit `fa2941d`.
- Previously recorded live validation on September 11, 2026: local and Runtime runs returned HTTP 200 and `payment_made: true` with one payment attempt and a 0.002 testnet USDC session debit. Both paths rejected a 0.0001 session budget without reducing the session balance.


01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_payment_agent.py
```diff
@@ -0,0 +1,174 @@
...
+        tools=[payment_tool],
+    )
+    print(f"Created OpenAI agent with name: {openai_agent.name}")
+    return openai_agent
+
+
+def run_agent(agent: Agent, prompt: str) -> str:
+    """Run the agent and return its final output."""
+
+    return str(Runner.run_sync(agent, prompt))
```
rajuans, line 149: **run_agent returns `str(Runner.run_sync(...))` instead of `.final_output`.** `Runner.run_sync` returns a `RunResult` object, and `str(result)` produces a verbose dump (e.g. `RunResult:\n- Last agent: ...\n- Final output (str): ...`) rather than the agent's actual answer. The runtime then returns `{"result": "<that dump>"}`, so the demo output is garbled. Use `Runner.run_sync(agent, prompt).final_output`.

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_x402_tool.py
```diff
@@ -0,0 +1,139 @@
...
+        response = httpx.request(method, url, timeout=30)
+        if response.status_code != 402:
+            return json.dumps(
+                {
+                    "status_code": response.status_code,
+                    "body": response.text,
+                }
+            )
+
+        client_token = str(uuid.uuid4())
```
rajuans, line 96: **`client_token` is generated once outside the retry loop and reused for every `generate_payment_header` attempt.** Because the client token is the idempotency key, each subsequent loop iteration returns the same (deduplicated) proof rather than a fresh payment. If the merchant returns 402 again, the loop burns all `MAX_PAYMENT_ATTEMPTS` producing identical results and never actually retries a distinct payment. Consider generating a fresh `client_token` per attempt (or documenting the intended single-payment idempotency and dropping the loop).

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_payment_agent.py
```diff
@@ -0,0 +1,174 @@
...
+        session_budget=os.getenv("PAYMENT_SESSION_BUDGET", "1.00"),
+    )
+
+
+def build_model(
+    model_region: str = "us-east-1",
+    model_id: str = "openai.gpt-5.5",
+) -> OpenAIResponsesModel:
+    """Configure the OpenAI Agents SDK for GPT-5.5 on Amazon Bedrock."""
+    set_tracing_disabled(False)
```
rajuans, line 84: **`set_tracing_disabled(False)` enables OpenAI Agents SDK tracing, whose default exporter needs an `OPENAI_API_KEY` this Bedrock-only setup never provides.** Each run emits trace-export warnings on stderr (e.g. `OPENAI_API_KEY is not set, skipping trace export`), polluting the demo output. Since auth is to Bedrock, not OpenAI, prefer `set_tracing_disabled(True)`.

rajuans (commented):
Automated code review (recall-focused pass). 7 findings on the new OpenAI Agents SDK payment example — 3 correctness issues (top-ranked), plus SSRF-guard hardening, a misleading error message, a weak tool schema, and a duplicated-default cleanup. Inline comments below. Non-blocking; posted as COMMENT.

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_x402_tool.py
```diff
@@ -0,0 +1,139 @@
...
+import uuid
+from urllib.parse import urlparse
+
+import httpx
+from bedrock_agentcore.payments import PaymentManager
+
+MAX_PAYMENT_ATTEMPTS = int(os.getenv("X402_MAX_PAYMENT_ATTEMPTS", "5"))
+
+
+def _validate_url(url):
```
rajuans, line 16: **SSRF guard is subject to TOCTOU / DNS rebinding, and the reserved-range check is incomplete.** `_validate_url` resolves DNS here, but `httpx.request` re-resolves independently at request time — a hostname that resolves to a public IP during validation and to `169.254.169.254` / `10.x` / `127.x` on the second resolution bypasses the check. Also only `is_private`/`is_loopback`/`is_link_local` are tested, so `is_reserved`, multicast, IPv4-mapped IPv6, and CGNAT (100.64/10) addresses pass. Consider pinning the validated IP into the request (or a shared HTTP client with a resolver) and broadening the address checks.

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_x402_tool.py
```diff
@@ -0,0 +1,139 @@
...
+        region_name=region,
+    )
+
+    def x402_fetch(url, method="GET"):
+        """Fetch a URL and automatically settle an x402 payment when required."""
+        url_error = _validate_url(url)
+        if url_error:
+            return json.dumps({"error": url_error})
+        if not user_id:
+            return json.dumps({"error": "PAYMENT_USER_ID environment variable is required"})
```
rajuans, line 85: **Misleading error message.** It references a `PAYMENT_USER_ID` environment variable, but `user_id` is a closure parameter passed into `build_x402_fetch`, not read from any env var. If `user_id` is empty this sends the user to set a nonexistent variable; the real source is `USER_ID` in `.env` / the invoke payload.

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_payment_agent.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+
+
+def build_agent(
+    model: OpenAIResponsesModel,
+    x402_fetch: Callable[..., str],
+) -> Agent:
+    """Build one payment-enabled OpenAI agent."""
+
+    payment_tool = function_tool(x402_fetch)
```
rajuans, line 134: **`function_tool(x402_fetch)` wraps a function whose params (`url`, `method`) have no type annotations.** The OpenAI Agents SDK builds the tool JSON schema from the signature; without annotations the params get an untyped/loose schema (or schema generation fails on stricter versions), degrading tool-calling reliability. Add annotations: `def x402_fetch(url: str, method: str = "GET")`.

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_payment_agent.py
```diff
@@ -0,0 +1,174 @@
...
+        model_id=os.getenv("BEDROCK_OPENAI_MODEL_ID", "openai.gpt-5.5"),
+        paid_url=os.getenv(
+            "PAID_URL",
+            "https://x402-test.genesisblock.ai/api/market-news",
+        ),
+        session_budget=os.getenv("PAYMENT_SESSION_BUDGET", "1.00"),
+    )
+
+
+def build_model(
```
rajuans, line 79: **Model region/id defaults (`"us-east-1"`, `"openai.gpt-5.5"`) are duplicated across three places:** `load_config`, this `build_model` signature, and `openai_payment_runtime.py`'s `os.getenv` fallbacks. Changing the default in one spot silently diverges local vs. runtime behavior. A single shared constant would keep them in sync.

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_payment_agent.py
mccartnick: Fixed in fa2941dda1419faa180afbb950dab9069ddf703d. `run_agent` now returns `Runner.run_sync(...).final_output` directly. The regression test checks the returned answer, and live GPT-5.5 runs locally and on AgentCore Runtime version 5 returned the final answer without the `RunResult` dump.

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_x402_tool.py
mccartnick: Fixed in fa2941dda1419faa180afbb950dab9069ddf703d using the suggested single-payment approach: the tool generates one proof with one idempotency key and replays the GET once. The payment loop and `X402_MAX_PAYMENT_ATTEMPTS` setting are removed. A repeated 402, merchant error, or timeout after proof generation reports an unknown payment outcome (`payment_made: null`) instead of minting another payment or claiming that nothing was paid. The README documents checking session spend before retrying. Regression tests cover repeated 402, redirects, HTTP 500, timeouts, and budget rejection; live local and Runtime runs each completed with one payment attempt and a 0.002 testnet USDC session debit.

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_payment_agent.py
mccartnick: Fixed in fa2941dda1419faa180afbb950dab9069ddf703d: `build_model` now calls `set_tracing_disabled(True)`. The default OpenAI exporter is disabled for this Bedrock-only example; Runtime CloudWatch/OpenTelemetry instrumentation is configured separately. Added a regression test for the tracing setting and verified the live local run without missing-OpenAI-key trace-export warnings.

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_x402_tool.py
mccartnick: Fixed in fa2941dda1419faa180afbb950dab9069ddf703d. Every DNS answer must pass public-address validation; both the initial GET and the paid replay connect to the same validated numeric IP. The original Host header and TLS SNI/hostname verification are preserved. Redirects and environment proxies are disabled, and each request uses a fresh client so challenge cookies are not replayed. The guard also rejects credentials in URLs, malformed ports, scoped IPv6, multicast, reserved/non-global addresses, CGNAT, and IPv4-mapped IPv6. Regression tests exercise unsafe/mixed DNS answers and a rebinding scenario; live HTTPS payment requests succeeded locally and on Runtime with certificate verification enabled.

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_x402_tool.py
mccartnick: Fixed in fa2941dda1419faa180afbb950dab9069ddf703d. The error now says `user_id is required; set USER_ID in .env or user_id in the invoke payload`. A regression test verifies both supported configuration sources and that the nonexistent `PAYMENT_USER_ID` name is absent.

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_payment_agent.py
mccartnick: Fixed in fa2941dda1419faa180afbb950dab9069ddf703d. The tool now declares `url: str`, `method: Literal["GET"] = "GET"`, and a `str` return type, with argument descriptions. GET is also enforced in code because this tutorial only retrieves content. Tests inspect the actual `function_tool` JSON schema with both the minimum supported and current SDK dependencies, and live model tool calls succeeded.

01-features/08-agents-that-transact/00-getting-started/01-agents-payments-and-limits/openai_payment_agent.py
mccartnick: Fixed in fa2941dda1419faa180afbb950dab9069ddf703d. `DEFAULT_MODEL_REGION` and `DEFAULT_MODEL_ID` are defined once in `openai_payment_agent.py` and used by `load_config`, the `build_model` defaults, and the Runtime entrypoint. Regression tests check that local and Runtime paths agree both with defaults and with environment overrides.

rajuans (approved):

mvangara10 (commented):
Remove this file

mvangara10 (approved):

Issue #2013, opened 2026-08-30
Sample: attenu-guard authorization for Strands agents on AgentCore Runtime


rafaelasor:
Proposing a sample under `03-integrations/agentic-frameworks/strands-agents/authorization/`, a sibling to the existing `observability/` sample.

What it shows: an orchestrator delegating to a sub-agent (agents-as-tools) whose permissions are computed as a subset of the parent's, so a hand-off can only narrow; an out-of-authority tool call cancelled before the tool body runs, through Strands' `BeforeToolCallEvent` hook, no monkeypatching; and the run's decisions recorded in a hash-chained audit log that verifies offline. The offline run uses a scripted model and needs no AWS account; the AgentCore deploy commands are documented, not executed.

The enforcement comes from attenu-guard (Apache-2.0, `pip install 'attenu-guard[strands]'`), which I maintain. It is not part of the AWS SDKs.

Branch is ready on my fork. Per CONTRIBUTING I am opening this issue first; the PR follows and will reference it.

PR #2014, opened 2026-08-30
Add attenu-guard authorization sample for Strands on AgentCore Runtime

rafaelasor:
Adds `03-integrations/agentic-frameworks/strands-agents/authorization/`, a
Strands + AgentCore Runtime sample showing:
- an orchestrator delegating to a sub-agent (agents-as-tools) whose permissions
  are computed as the meet of its request and the parent's authority, so a
  hand-off can only ever narrow
- an out-of-authority tool call cancelled before the tool body runs, via
  Strands' `BeforeToolCallEvent` hook (no monkeypatching)
- the run's decisions recorded in a hash-chained, offline-verifiable audit log

Enforcement comes from `attenu-guard` (Apache-2.0, not part of the AWS SDKs).

How to run: `uv venv && uv pip install -r requirements-dev.txt`, then
`python local_run.py` (offline, scripted model, no AWS account) and
`python -m pytest`. The AgentCore deploy commands (`agentcore configure` /
`launch` / `invoke`) are documented in the README but not executed as part
of this PR.

What the reader sees: the sub-agent reads logs (permitted), then tries to
export findings to an external destination (not permitted) — the call is
denied before the tool body runs, and the offline run prints the denial plus
a verified evidence bundle.

Refs #2013 (opened first per CONTRIBUTING).

Disclosure: I maintain attenu-guard.

### DIFF
--- a/03-integrations/agentic-frameworks/strands-agents/authorization/README.md
+++ b/03-integrations/agentic-frameworks/strands-agents/authorization/README.md
@@ -1,0 +1,261 @@
+# Strands sub-agent permissions with AgentCore Runtime
+
+| Information | Details |
+|---|---|
+| Agent type | Synchronous |
+| Agentic framework | Strands (agents as tools) |
+| LLM model | Whatever the Runtime is configured with; the offline run uses a scripted model |
+| Components | AgentCore Runtime |
+| Example complexity | Easy |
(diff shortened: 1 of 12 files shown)


rafaelasor:
Open since 30 August with no reviewer and no assignee. Is there a different path for community samples under `03-integrations`, or someone I should ask directly?

PR #2078, opened 2026-09-16
feat(data-platforms/databricks): native TOKEN_EXCHANGE per-user sample with preflight check

antonyprasad-db:
## What this adds

A native `TOKEN_EXCHANGE` path for per-user (on-behalf-of) access from an AgentCore agent to a Databricks Managed MCP server, alongside the existing interceptor sample.

Gateway now supports the `TOKEN_EXCHANGE` grant type natively, so the identity exchange no longer has to happen in a Lambda. The configuration is three objects and no code: a `CUSTOM_JWT` inbound authorizer, a credential provider carrying `onBehalfOfTokenExchangeConfig`, and an MCP target.

## Why a preflight script is included

A misconfigured on-behalf-of target does not necessarily fail. It can return HTTP 200 with no error and run every query as the shared service principal, which in a governed environment is worse than an outage: someone following the documented configuration believes per-user delegation is active while the audit trail attributes nothing to the person who asked.

Nothing in the API or the console reports which identity will actually arrive at Unity Catalog, so `check_obo_identity.py` asks the only question that settles it — does `current_user()` come back as a human or as the service principal? It creates and changes nothing, and it maps each known failure string to a cause, including one that reads like a service-side restriction but is in fact the caller's own IAM policy.

## Prerequisites now written down

These are not currently documented anywhere and each one produces a distinct, non-obvious failure:

- an **account-level** Databricks OAuth federation policy (a workspace-level policy is not sufficient)
- **end-user workspace membership** — without it the exchange resolves the identity correctly and then fails, so identity mapping is working and provisioning is the gap
- the service principal's `workspace-access` and `databricks-sql-access` entitlements, and re-minting its token afterwards
- `mcp.mcpServer.listingMode: DYNAMIC` on Managed MCP targets
- a gateway execution role policy that is **not** pinned to a single Region

## Verification

Verified end to end against a Databricks workspace:

- a `TOKEN_EXCHANGE` target returned `current_user()` as the **end user**
- a `CLIENT_CREDENTIALS` control against the same workspace and endpoint returned the **service principal**, so the identity mode is the only difference
- a byte-identical configuration in a Region where the public-client exchange was not enabled did not complete, which is why the README documents how to request it

The reason no client authentication may be presented was measured directly against the same workspace with the same subject token: with client authentication the exchange returns `400 invalid_grant`, because Databricks then looks for a service-principal federation policy rather than the account user policy; without it, the returned token's subject is the end user. Databricks advertises `none` in `token_endpoint_auth_methods_supported`.

## Change to the existing interceptor sample

Two edits, six lines. A pointer to the native path, and a correction to the sentence stating that token exchange is a grant type Gateway does not support natively — accurate when written, no longer so. The interceptor sample remains the working approach where the native path is not yet enabled, so nothing is removed or deprecated.

## Testing

```
python -m unittest discover -v    # 46 tests, standard library only
```

`ruff check` and `ruff format --check` are clean against the repository configuration. The tests cover the identity and failure classification, response parsing including multi-frame SSE bodies, and request construction; each was verified to fail when the corresponding behaviour is broken.


antonyprasad-db:
@Robobc tagging you as the author of the interceptor sample, since this is the collapse into native config you expected once `TOKEN_EXCHANGE` reached production — the exchange now carries the inbound user token and Unity Catalog sees the end user, so the interceptor Lambda is not needed where the native path is enabled.

I have deliberately left `databricks-dbsql-per-user-delegation` in place rather than deprecating it, since it remains the working approach for accounts where the public-client exchange is not yet enabled. The only change there is a pointer to the new sample and a correction to the sentence saying Gateway does not support the grant natively.

The piece I would most value your read on is `check_obo_identity.py`. A misconfigured on-behalf-of target can return 200 and still run as the service principal, and nothing in the API reports which identity arrived, so the script classifies the principal that actually reached Unity Catalog and maps each failure string to its cause. If any of those strings are ones you would phrase differently, or if there is a cause I have mapped wrongly, I would rather fix it here than have customers chase it.

I cannot add reviewers on this repo, so tagging rather than assigning.

Issue #2093, opened 2026-09-18
03-bidirectional-streaming: run the five samples on AgentCore Runtime V2 (platformVersion)


deepaxs:
Migrate the five samples under 01-features/02-host-your-agent/01-runtime/03-advanced/03-bidirectional-streaming — 01-bedrock-sonic-ws, 02-strands-ws, 03-langchain-transcribe-polly-ws, 04-pipecat-sonic-ws and 05-bidirectional-streaming-webrtc — to AgentCore Runtime V2 by setting platformVersion="V2" on create_agent_runtime.

Additionally, two per-existing defects stop these samples working regardless of platform version and had to be fixed before V2 could be evaluated: bedrock:InvokeModelWithBidirectionalStream is missing from agent_role.json , and unbounded dependencies in samples 01 and 05 resolve to a version where the container fails at import.

**Scope**
deploy.py  plus a new deploy.py , agent_role.json , four server.py files, sample 05's audio.py and nova_sonic.py, three requirements.txt, sample 05's two IAM policy files, and six READMEs.

All five samples were validated by holding a real spoken conversation with each one.

evandrofranco:
Solved in this PR: https://github.com/awslabs/agentcore-samples/pull/2094

PR #2094, opened 2026-09-18
feat(runtime/bidirectional-streaming): run the five samples on AgentCore Runtime V2

deepaxs:
# Amazon Bedrock AgentCore Samples Pull Request

> [!IMPORTANT]  
> 1. We strictly follow a issue-first approach, please first open an [issue](https://github.com/awslabs/amazon-bedrock-agentcore-samples/issues) relating to this Pull Request.
> 2. Once this Pull Request is ready for review please attach `review ready` label to it. Only PRs with `review ready` will be reviewed.

**Issue number**:  #2093

```
Changes the five 03-bidirectional-streaming samples to deploy on AgentCore Runtime V2,
because platformVersion="V2" also requires adapting each agent to the snapshot model, and
two pre-existing defects had to be fixed before V2 could be evaluated at all.
```

## Summary

> **For reviewers.** The change itself is one field, `platformVersion="V2"`. Most of this PR is
> its consequence: state that was safe to capture at container startup is not safe to capture
> into a snapshot. If you read one section, read
> [For maintainer review: how the container image gets built](#for-maintainer-review-how-the-container-image-gets-built)
> -- it is the only thing here I need a decision on. The rest is the snapshot adaptation, plus
> two defects that stopped these samples working on any platform version. These are also the
> first **container-based** samples to move to V2; the four converted so far use direct code
> deployment and build no image.

All five voice-agent samples under
`01-features/02-host-your-agent/01-runtime/03-advanced/03-bidirectional-streaming/` now
deploy to **AgentCore Runtime V2** by setting `platformVersion="V2"` at runtime creation.

V2 starts the container once while the runtime is being created, snapshots it, and restores
that snapshot for new execution environments. Most of this PR is the consequence of that
change rather than the field itself: state that was safe to capture at startup on the
previous platform is not safe to capture into a snapshot, and two samples had dependency
declarations loose enough that a container could fail to start at all, independent of either
platform version.

All five samples were validated by holding a real spoken conversation with each one.

### Deploy tooling

Neither the starter toolkit nor the AgentCore CLI exposes `platformVersion`, so the deploy
path builds the image with the existing managed build service and then creates the runtime
through a direct `bedrock-agentcore-control` call.

| File | Change |
|------|--------|
| `utils/deploy.py` | build the ARM64 image via CodeBuild, then create or update the runtime with `platformVersion="V2"`; poll `GetAgentRuntime` for readiness; pass the deploy region into the container |
| `05-bidirectional-streaming-webrtc/deploy.py` | **new file.** Sample 05 previously had no deploy script and its README drove CLI steps that cannot set `platformVersion`. Creates the execution role with its policies attached, builds the image, and creates the runtime in VPC network mode |
| `utils/requirements.txt` | add the starter toolkit; raise the boto3 floor to 1.43.95, which accepts `platformVersion` |
| `utils/start_client.sh` | drop two folder names that do not exist; point at `deploy.py` instead of a `setup.sh` that was never present |

`CreateAgentRuntime` and `UpdateAgentRuntime` accept `platformVersion` but do not echo it
back, so both scripts confirm it with `GetAgentRuntime` after the call. Failure detection
matches on a `FAILED` suffix rather than enumerating statuses, so it covers both
`CREATE_FAILED` and `UPDATE_FAILED`.

### Execution role permissions

| File | Change |
|------|--------|
| `utils/agent_role.json` | add `bedrock:InvokeModelWithBidirectionalStream` |

`bedrock:InvokeModelWithBidirectionalStream` was **absent entirely**. The role granted
`bedrock:InvokeModel` and `bedrock:InvokeModelWithResponseStream`, neither of which is used
by two-way audio streaming. Confirmed by a real voice session: every conversation failed
with an authorisation error until this was added.

### Making the samples deployable and startable at all

Two things had to be fixed before Runtime V2 could be evaluated on these samples, because
they stopped a container from ever reaching a healthy state in the first place.

**Dependency pins.** `01-bedrock-sonic-ws` and `05-bidirectional-streaming-webrtc` declared
dependencies with no upper bound. `aws-sdk-bedrock-runtime` 0.8.0 renamed
`BedrockRuntimeClient` to `AsyncBedrockRuntimeClient` and `Config` to
`AsyncBedrockRuntimeConfig`; `smithy-aws-core` 0.8.0 removed `create_default_chain`. Sample
01's `s2s_session_manager.py` imports the first two; sample 05's `nova_sonic.py` imports all
three. An unbounded install resolves to a version where the module fails at import and the
container exits immediately -- confirmed by a real deployment, which logged the `ImportError`
and never reached healthy.

| File | Change |
|------|--------|
| `01-bedrock-sonic-ws/websocket/requirements.txt` | cap `aws_sdk_bedrock_runtime` below 0.8.0 |
| `05-.../agent/requirements.txt` | cap the Nova Sonic SDK and its three smithy dependencies below their breaking releases; cap `aiortc` and `av` to the tested majors (`aiortc` 1.15.0, `av` 17.1.0 -- a pip dry-run against these bounds resolves to exactly those versions, and pip's own metadata fetch shows `av` 18.x already exists on PyPI) |

This failure is a Python import error that occurs before any AgentCore Runtime code
executes, so it is independent of platform version -- the class rename breaks the same way
whether the container boots fresh or resumes from a snapshot. It is included here rather than
filed separately because every sample in this PR ships on V2, and a customer following any of
these READMEs should get a container that starts.

Sample 02 is unaffected because it happens to pin an exact version already. Samples 03 and 04
do not import these packages.

### Adapting the agents to the snapshot model

Anything an application captures during startup is frozen into the deploy-time snapshot and
restored later, potentially hours afterwards.

| File | Change |
|------|--------|
| `01-bedrock-sonic-ws/websocket/server.py` | resolve AWS credentials per session |
| `02-strands-ws/websocket/server.py` | resolve AWS credentials per session |
| `03-langchain-transcribe-polly-ws/websocket/server.py` | resolve AWS credentials per session |
| `04-pipecat-sonic-ws/websocket/server.py` | resolve AWS credentials per session |
| `05-.../agent/audio.py` | build the FFmpeg resampler and silence frame per session instead of at module import |
| `05-.../agent/nova_sonic.py` | construct the resampler in `run_session` and pass it into `convert_to_16khz` |

The credential change fixes a `403 ExpiredTokenException` observed on a session restored
several hours after deployment. Startup-captured credentials had expired, and the
application's own background refresh loop does not resume after a snapshot restore.

The `audio.py` change fixes a **silent** failure, the least obvious consequence of the snapshot
model: the `av.AudioResampler` built at module import was captured in the snapshot and restored
in an unusable state, returning correctly sized audio frames containing silence, so the sample
deployed, connected and streamed audio while the model returned nothing at all -- no error, no
warning, no timeout. Isolated against a control runtime identical in image, execution role,
subnet, security group and environment variables and differing only in platform version, which
transcribed and answered normally; moving construction into the session resolved it.

As a side effect this also fixes a latent bug unrelated to the platform: a single resampler
shared across concurrent peer connections was never correct.

### Sample 05 IAM policy templating

| File | Change |
|------|--------|
| `05-.../kvs-iam-policy.json` | region and channel name become placeholders |
| `05-.../bedrock-iam-policy.json` | region becomes a placeholder |

The new `deploy.py` attaches these policies automatically, where previously the README told
the reader to edit them by hand. That automation exposed a problem: only the account ID was
being substituted, so the `--channel-name` flag produced a role scoped to a channel the agent
never uses. Confirmed by policy simulation -- the default channel evaluated to `allowed`, any
other channel to `implicitDeny`.

Rendering the templated files at default values produces a policy **byte-identical** to what
was deployed and tested, so behaviour out of the box is unchanged.

### Documentation

| File | Change |
|------|--------|
| `03-bidirectional-streaming/README.md` | rewritten: all five samples listed, Runtime V2 explained once, a broken pointer removed |
| `01`, `02`, `03`, `04`, `05` sample READMEs | one line each stating V2 and the boto3 requirement, linking to the directory README; sample-specific notes retained |

Two corrections to the directory README are worth calling out.

It carried a section titled "Why This Example Doesn't Include Deploy Scripts" -- no longer
true, since the directory contains `utils/deploy.py` and now a second deploy script. It also
pointed readers at `01-AgentCore-runtime/06-bi-directional-streaming/` for "four complete,
working examples". That path does not resolve from the repository root (it is missing a
`06-workshops/` prefix), the directory here holds five samples rather than four, and the table
omitted sample 05.

The four sample READMEs previously instructed readers to `cd` into a **different copy** of
these same samples elsewhere in the repository. That path did resolve, so anyone following the
documentation from the repository root would have deployed the other copy. They now refer to
their own directory.

### Repository hygiene

| File | Change |
|------|--------|
| `.gitignore` | ignore `**/setup_config.json`, which the deploy scripts generate and which contains an AWS account ID |

## For maintainer review: how the container image gets built

These are the first **container-based** samples to move to Runtime V2. The four converted so far
-- the three under `01-http-protocol` and `03-advanced/07-persistent-filesystems` -- all use
direct code deployment, so none of them builds an image. That difference is what this section is
about, and it is the one decision here I would rather have confirmed than merged on my judgement.

**What these samples do.** `utils/deploy.py` and `05-.../deploy.py` build the ARM64 image through
the starter toolkit's CodeBuild path, calling `_execute_codebuild_workflow(..., ecr_only=True)`,
then create the runtime with a direct `bedrock-agentcore-control` call carrying
`platformVersion="V2"`. The create half is unavoidable, since no tooling exposes that field. The
build half uses a private toolkit function, which is the part worth your attention.

**The alternatives, in the order they were evaluated and why each was rejected:**

- **The toolkit's public API.** `launch()` builds and creates the runtime in a single call. The
  only point separating the two is the private function above, so there is no public build-only
  entry point to use instead.
- **The `agentcore` CLI.** It does have a build-only command (`agentcore package`), so the gap is
  not the one it first appears to be. The blocker is that the CLI does not pass `platformVersion`:
  absent from the shipped code of the latest public release, 0.30.0 (checked 2026-09-18). The
  field has since landed in the JS SDK (`@aws-sdk/client-bedrock-agentcore-control` 3.1134.0), but
  0.30.0's `npm-shrinkwrap.json` pins 3.1107.0, which does not carry it -- so the CLI cannot set
  V2 even indirectly. It is also project-scoped, expecting an `agentcore.json` + CDK layout these
  samples do not have.
- **A local `docker buildx` ARM64 build.** Removes the toolkit dependency outright, but requires
  every reader to run Docker with ARM64 emulation, giving up the no-local-Docker property that
  makes the CodeBuild path worth having in a sample.
- **Reimplementing the CodeBuild orchestration on boto3.** Also removes the private import, at the
  cost of an ECR repository, a CodeBuild project, a service role and the polling around them --
  added to a file whose purpose is to teach Runtime V2.

**How the risk is contained.** `utils/requirements.txt` pins
`bedrock-agentcore-starter-toolkit>=0.3.12,<0.4.0`, the range this was tested against, so a
routine toolkit release cannot change or remove the function underneath the script. The call site
names the symbol in a comment so the next person knows what to re-check before widening the bound.

**The question for you:** is a pinned private import acceptable in a public sample, or would you
prefer the boto3 reimplementation with its extra build plumbing? Either is fine by me and I am
happy to switch. Separately, a public build-only parameter on `launch()`, or the CLI passing
`platformVersion`, would remove the need entirely and let `deploy.py` drop one import with no
other change -- proposed as a separate request to the toolkit maintainers.

**One thing I did not test:** whether `agentcore package` on its own produces an ECR image or only
stages artifacts for `deploy`. It does not change the conclusion, because the create step blocks
the CLI either way, but I would rather flag it than imply I ruled it out.

## Testing

Every sample was deployed and driven with a real microphone. This mattered: several defects
were invisible from deployment status, and each sample reached a healthy state and served
health checks while being non-functional.

| Sample | Result |
|--------|--------|
| 01 | voice conversation, barge-in, tool use |
| 02 | voice conversation across four MCP gateways, memory recall, per-session credential refresh observed in logs |
| 03 | Transcribe to LLM to Polly pipeline, three tool calls returning expected values |
| 04 | voice conversation with audible responses, all three tool calls confirmed in runtime logs |
| 05 | voice conversation after the resampler fix, at parity with the control runtime |

Static verification: `ruff check` and `ruff format --check` pass on the changed Python files
with zero new findings against `HEAD`, the JSON policy files parse, the cross-README links
resolve, and a pip dry-run confirms the new `aiortc`/`av` bounds resolve to the versions used
in the tested image.

## New dependencies

`bedrock-agentcore-starter-toolkit` is added to `utils/requirements.txt`, used only at deploy
time to build the container image through CodeBuild. No new runtime dependencies are
introduced; the dependency changes in this PR are upper bounds on existing packages, not new
additions.

### User experience
**Before.** The samples deploy to the previous platform. Following any of the four sample READMEs from the repository root cds the reader into a different copy of these samples elsewhere in the repo, so the code they deploy is not the code they are reading. A fresh build of samples 01 and 05 produces a container that exits at import. Once running, every voice session fails authorisation, because the execution role lacks the only Bedrock action these samples use.

**After.** All five deploy on Runtime V2, with the field confirmed via get_agent_runtime. Each README points at its own directory. Containers start, and a spoken conversation works end to end on every sample — including sessions restored from a snapshot hours after deployment.

### Note on overlapping work

PR #2060 (open) pins `mcp<2` in `02-strands-ws/mcp/requirements.txt`. Same directory, different file from anything changed here, so no conflict.

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [x] Add your name to [CONTRIBUTORS.md](https://github.com/awslabs/amazon-bedrock-agentcore-samples/blob/main/CONTRIBUTORS.md)
* [ ] Have you documented **Introduction**, **Architecture Diagram**, **Prerequisites**, **Usage**, **Sample Prompts**, and **Clean Up** steps in your example README?
  <br/>Partially, and unchanged by this PR. All five have Introduction (as `Overview`), Usage and Clean Up, and samples 01-04 embed architecture diagrams from `../assets/`. None has a `Prerequisites` or `Sample Prompts` section — and none did before this PR either, so this change neither adds nor removes them. Happy to add `Prerequisites` here if you would like, since this change does introduce one (boto3 >= 1.43.95).
* [x] Changes are documented

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/amazon-bedrock-agentcore-samples/blob/main/LICENSE).


evandrofranco (dismissed):

evandrofranco (approved):
ok

PR #2108, opened 2026-09-23
payment manager default for --auto-payment 

mvangara10:


01-features/08-agents-that-transact/00-getting-started/00-setup-agentcore-payments/README.md
```diff
@@ -135,6 +135,12 @@ cd PaymentSetup
 agentcore add payment-manager --name MyPaymentManager --auto-payment true --default-spend-limit 1.00
 ```
 
+> **`add payment-manager` defaults to auto-payment ON with a $10.00 spend cap.** If you omit
+> `--auto-payment` and `--default-spend-limit`, the CLI defaults to `--auto-payment true` and
+> `--default-spend-limit 10.00`. This command explicitly lowers that cap to `1.00` for the tutorial —
+> when adapting this command for your own project, set `--default-spend-limit` deliberately rather
+> than relying on the $10 default.
+
```
Khusbu: Can you add PR description for the change?

Khusbu: How did we arrive at $10 default spend limit? What if an agent runs the skill and spends $10 from customer's wallet (mainnet) without them knowing, $10 dollar might be a higher amount for some? Should we reduce the default spend limit to $1 to be on the safe side?

akshseh (approved):

evandrofranco (approved):
ok

PR #2110, opened 2026-09-23
Rename the mastering agent to the delivery agent in gpu-music-production-agent

rmncardoso:

```
Renames the mastering agent to the delivery agent throughout
02-use-cases/02-workflow-automation-agents/gpu-music-production-agent, adopting the
term the sample already used for this concept rather than introducing a new one.
```

The sample already spoke of "delivery" for exactly this stage: the agent is handed a
**delivery target**, the compliance agent ran **delivery QC** against `delivery.json`,
and its system prompt was already given "a rendered mix and a delivery target". The
rename makes the vocabulary internally consistent.

| Before | After |
|---|---|
| `mastering_agent.py` | `delivery_agent.py` |
| `Dockerfile.mastering` | `Dockerfile.delivery` |
| `MasteringPlan` | `DeliveryPlan` |
| `master.wav` | `delivery.wav` |
| `mastering.json` / `mastering.md` | `delivery.json` / `delivery.md` |
| `master_is_stale` | `delivery_is_stale` |
| `is_master` | `is_delivery` |
| `delivery_qc()` | `technical_qc()` |
| `AGENT_NAME = "mastering"` | `"delivery"` |
| `MASTERING_MODEL_ID` | `DELIVERY_MODEL_ID` |

Two choices worth calling out:

- **`delivery_qc()` became `technical_qc()`** so it pairs with `similarity_screen()`.
  Those are the compliance agent's two independent checks, and keeping the old name
  would have read as though the function belonged to the delivery agent.
- **The system prompt keeps its domain signal without the word** — "an audio engineer
  who prepares finished mixes for release delivery", not "a delivery engineer", which
  means something else entirely. This string steers model behaviour, so it was
  re-validated on a live run rather than assumed (see Testing).

### User experience

**Breaking for an existing deployment.** The agent key in `deployment_state.json` moves
from `"mastering"` to `"delivery"`, and that key drives `Dockerfile.{agent}`, the ECR
repository (`music-production/delivery-agent`), the AgentCore runtime name
(`music_production_delivery_*`) and the Strands session directory
(`.sessions-delivery`). A deployment created before this change needs a fresh
`deploy.py` rather than `update.py`. Nothing else about the workflow changes.

The run output reads the same, with the step renamed:

```
  4. delivery (real DSP, verified by measurement)
    read     : composition.wav  <- written by another agent
    before    20.062s  48000Hz  2ch  -6.87 LUFS  peak 0.2 dBTP  LRA 1.45 LU
    after     20.062s  48000Hz  2ch  -14.0 LUFS  peak -3.84 dBTP  LRA 1.47 LU
    targets  : loudness met, true peak held

  -- remediation happened, so the delivery is stale: re-preparing --
  6. re-prepare the delivery for the replacement
  7. re-screen the new delivery
```

Artifacts in `runs/<track>/` are now `composition.wav`, `delivery.wav`, `delivery.md`,
`compliance.md` and `compliance.json`.

### Testing

Deployed from scratch and run end to end on a live `g6.xlarge` in `us-east-2`, then torn
down and verified clean.

- **All seven steps on one instance** (`ip-172-31-15-195`), so collocation still holds
  across the cross-agent callback; remediation fired and steps 6–7 ran.
- **Confirmed by measurement, not by the agents' own reports.** The downloaded audio was
  re-measured locally with the sample's own DSP and matched exactly: `delivery.wav` at
  29.907 s, −14.01 LUFS, −1.0 dBTP, 7.87 LU, with clipped runs going 13 → 0.
- **Response schema verified renamed**: `technical_qc` present, `delivery_qc` absent,
  `delivery_is_stale` present. S3 keys write `delivery.wav` / `delivery.md`.
- **Model behaviour unchanged under the reworded prompt**: it caught a DC offset at
  0.001917, bypassed the compressor with a stated reason (loudness range 7.63 LU is
  healthy), and made three conservative EQ moves each tied to a measurement.
- **Every script exercised**: `deploy.py`; `invoke.py --with-catalogue` and `--resume`;
  `update.py delivery` (resolved `Dockerfile.delivery` and the new ECR repo, shipping
  only that agent — the other two stayed at version 1); `update.py --all`;
  `cleanup.py` (full teardown, 0 instances and 0 orphaned volumes afterwards).
- `ruff check` and `ruff format --check` pass at line-length 120; images 392.6 / 392.5 MB
  against the 2 GB cap, compliance zip 82.3 MB / 235.6 MB against 250 / 750.


evandrofranco (approved):
OK

